PR #1, opened 2022-01-25
Bring code from `conda/conda`

jaimergp:
We started this work as a feature branch in `conda/conda`, mainly through this [PR](https://github.com/conda/conda/pull/10881) (and its many children PRs). As we get closer to the release dates, we need to prepare that code for opt-in distribution, which will involve a separate package our users can install next to `conda` in their `base` environment.

In this PR, we will:

- [X] bring the needed code from the aforementioned PR and adjust as necessary
- [ ] add the necessary metadata bits to make it conda-installable
- [ ] add CI workflows

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,74 @@
 # conda-libmamba-solver
+
+<!-- TODO: Needs to be updated for the new namespaces -->
+
+## Trying solvers
+
+The new libmamba integrations are experimental, but you can get a taste of how they are working
+so far by following these instructions.
+
+1. Clone `conda/conda` and checkout the branch you want to try. `poc-libsolv` is the "stable
+feature branch", but there might be other ones that are interesting to test. Check the opened
+PRs to see.
+
+```
+git clone https://github.com/conda/conda
+cd conda/
+git checkout poc-libsolv
+```
+
+2. Build the Docker image:
+
+```
+docker compose build --no-cache interactive
+```
+
+3. Run the interactive console. This will preinitialize the image for `conda` in `dev` mode,
+which means that:
+    * The `conda` repository will be mounted to `/opt/conda-src``
+    * The default `base` environment will run off this development install of `conda`!
+
+```
+docker compose run interactive
+```
+
+4. Now you can experiment different things. `--dry-run` is specially useful to check how different
+solvers interact. The main switch you need to take care of is the _solver logic_ option:
+
+```bash
+# Using default (classic) solver
+$ conda create -n demo scipy --dry-run
+# This is equivalent
+$ CONDA_SOLVER_LOGIC=classic conda create -n demo scipy --dry-run
+# Using original libmamba integrations
+$ CONDA_SOLVER_LOGIC=libmamba conda create -n demo scipy --dry-run
+# Using refactored libmamba
+$ CONDA_SOLVER_LOGIC=libmamba2 conda create -n demo scipy --dry-run
+```
+
+> `mamba` is also available in case you want to compare our integrations with the original Mamba
+> project: `mamba create -n demo scipy --dry-run`
+
+5. Use `time` to measure how different solvers perform. Take into account that repodata
+   retrieval is cached across attempts, so only consider timings after warming that up:
+
+```bash
+# Warm up the repodata cache
+$ conda create -n demo scipy --dry-run
+# Timings for original solver
+$ time env CONDA_SOLVER_LOGIC=classic conda create -n demo scipy --dry-run
+# Timings for libmamba integration
+$ time env CONDA_SOLVER_LOGIC=libmamba2 conda create -n demo scipy --dry-run
+```
+
+> `conda create` commands will have similar performance because it's a very simple action! However,
+> things change once you factor in existing environments. Simple commands like `conda install scipy`
+> show ~2x speedups already.
+
+6. If you need extra details on _why_ solvers are working in that way, increase verbosity. Output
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1 +1,80 @@
...
+docker compose run interactive
+```
+
+4. Pip install this repository, with no deps:
+
+```
+python -m pip install --no-deps https://github.com/conda-incubator/conda-libmamba-solver/archive/main.tar.gz
+```
+
+5. Now you can experiment different things. `--dry-run` is specially useful to check how different
```
beeankha, line 41: "...experiment with different things." (add "with")

conda_libmamba_solver/libmamba_utils.py
```diff
@@ -0,0 +1,320 @@
+# Copyright (C) 2019, QuantStack
+# SPDX-License-Identifier: BSD-3-Clause
+
+# TODO: Temporarily vendored from mamba.utils v0.19 on 2021.12.02
+# Decide what to do with it when we split into a plugin
```
jezdez, line 5: Just making sure we're comfortable adding this at the moment, do we need anything specific from this, or all of it?

conda_libmamba_solver/state.py
```diff
@@ -0,0 +1,1404 @@
...
+from conda.models.prefix_graph import PrefixGraph
+from conda.core.index import _supplement_index_with_system
+from conda.core.prefix_data import PrefixData
+from conda.core.solve import get_pinned_specs
+
+logger = logging.getLogger(__name__)
+
+
+class TrackedMap(MutableMapping):
+    # TODO: I am sure there's a better place for this object; e.g. conda.models
```
jezdez, line 92: Should we add a `conda_libmamba_solver/models.py`?

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+requires = [
+  "setuptools >= 42",
+  "wheel",
+  "setuptools_scm[toml]>=3.4"
+]
```
jezdez, line 6:
```suggestion
requires = [
  "setuptools >= 45",
  "wheel",
  "setuptools_scm>=6.2"
]
```
setuptools_scm has gotten an update to not need the toml extra anymore but relies on the tomli package directly.

setup.cfg
```diff
@@ -0,0 +1,27 @@
+[metadata]
```
jezdez, line 1: What do you think about porting this to a `pyproject.toml` instead and lean fully into current Python packaging standards? Would go well with our plans around conda plugins via pyproject.toml.

setup.cfg
```diff
@@ -0,0 +1,27 @@
...
+name = conda-libmamba-solver
+version = 0.1.0
```
jezdez, line 3: Just generally when using setuptools_scm the version is not needed.

jezdez (changes requested):
No surprises, and just a few nits, mostly around packaging and utility code. 🌟

conda_libmamba_solver/__init__.py
```diff
@@ -0,0 +1,10 @@
+
+from .libmamba import LibMambaSolver
+from .libmamba2 import LibMambaSolver2
+
+
+def get_solver_logic(key=None):
+    return {
+        "libmamba": LibMambaSolver,
+        "libmamba2": LibMambaSolver2,
```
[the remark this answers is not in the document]
jaimergp: Yes. For now I am keeping both around and accessible, but for the release we might want to only offer one. Or at least name them in a way that is more understandable:

* `libmamba2` would become `libmamba`
* `libmamba` would become `libmamba-draft` or something like that

conda_libmamba_solver/libmamba_utils.py
jaimergp: I only kept the parts we need. If we don't ship the current `libmamba` we can remove a few more.

conda_libmamba_solver/state.py
jaimergp: Good idea!

pyproject.toml
jaimergp: Awesome! I was a bit unsure about these bits. First project I create with PEP517 in mind!

setup.cfg
jaimergp: All for it. I didn't understand why _two_ files were needed, but the PyPA example and others I've seen live had all of it 🤷

conda_libmamba_solver/__init__.py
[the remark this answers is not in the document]
jaimergp: I added this change now.

conda_libmamba_solver/libmamba_utils.py
jaimergp: Notice that we had to add some small fixes here for our own purposes, which might have to be upstreamed.

conda_libmamba_solver/__init__.py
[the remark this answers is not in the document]
jezdez: Yeah, good idea!

conda_libmamba_solver/libmamba_utils.py
jezdez: Assuming we remember which things to upstream, I think Wolf would appreciate it.

jaimergp:
Ready to merge @jezdez

jezdez (approved):
Let's get the ball rolling.

PR #8, opened 2022-02-15
Change stderr capture from pipe to tempfile

jaimergp:
Repodata downloads hangs on Windows if `stderr` is captured via `utils.CapturedDescriptor`. 

I have tried threaded vs not threaded, or even disabling progress bars by setting `CI=1` as an environment variable. It looks like duplicating that stream just blocks that call.

@wolfv - is it possible to have the libraries log to a path of our choice? I think forwarding through `logging` is not feasible on Windows, but at least we can redirect the logged lines from libmamba to a file there.

### DIFF
--- a/tests/test_experimental.py
+++ b/tests/test_experimental.py
@@ -54,12 +54,11 @@ def test_logging():
     else:
         pytest.fail("Could not find logfile path in outout")
 
-    if not on_win:  # file logging on windows is temporarily disabled :(
-        with open(logfile_path) as f:
-            log_contents = f.read()
-            assert "conda.conda_libmamba_solver" in log_contents
-            assert "solver started" in log_contents
-            assert "choice rule creation took" in log_contents
+    with open(logfile_path) as f:
+        log_contents = f.read()
+        assert "conda.conda_libmamba_solver" in log_contents
+        assert "solver started" in log_contents
+        assert "choice rule creation took" in log_contents
 
 
 def test_cli_flag_in_help():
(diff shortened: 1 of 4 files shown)


tests/test_experimental.py
```diff
@@ -96,9 +96,11 @@ def cli_flag_and_env_var_settings():
     ]
     return tests
 
+
 @pytest.mark.parametrize("name, command, env, solver", cli_flag_and_env_var_settings())
 def test_cli_flag_and_env_var(name, command, env, solver):
         stdout = check_output(command, env=env, stderr=STDOUT, universal_newlines=True)
+        print(stdout)
```
jezdez, line 103: Nit

jezdez (approved):
This is an acceptable work-around, but we should fix this in a next version.

jaimergp:
To be clear, we still have file logging on Windows, but it will lack the internal `libsolv` details.

wolfv:
Hmm, I would like to understand why this has changed ... it might be related to the new progress bars but I can't really see how.

wolfv:
@jaimergp any chance you have a minimal reproducer? I'd be happy to investigate this further in mamba. We added some Windows specific functions to read information about the attached terminal but I don't think that they'd block like this. 
It's possible that there is a threading issue with the new progress bars -- that would be nice to resolve.

jaimergp:
Hm, I don't think it has changed. I can reproduce with 0.19, 0.20 and 0.21. We didn't see it on CI earlier because I guess an interactive console is required to trigger it. I'll try to come up with a reproducer involving `os.dup`, `os.dup2` and friends.

jaimergp:
We might need to opt for a different way across all OS. This won't do any good down the line and it's only temporary anyway.

I've discovered that if an exception happens during the capture, since we are capturing `stderr`, the traceback won't be printed :D Conda error reporting will fetch it and ask if we want to submit the report, thus "hanging". Might explain hangs in macOS too.

On Windows the issue seems to be that too much output is hogging the pipe, but it's tricky to reproduce. Also, the standard streams are somehow wrapped with colorama stuff, so I had to write the sentinel with `os.write` instead.

@beckermr - this might explain your issues with wurlitzer too.

jaimergp:
This seems to be working :D

jaimergp:
@kenodegard has confirmed this works and fixes issues on macOS. I checked locally on Windows and same thing. Tests are passing... I guess we should merge!

PR #9, opened 2022-02-15
Fix the condition for canary builds.

jezdez:
I noticed that the canary builds weren't building anymore, trying to understand what's going on here.

### DIFF
--- a/.github/workflows/builds-canary.yaml
+++ b/.github/workflows/builds-canary.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
-    if: github.event.workflow_run.conclusion == 'success'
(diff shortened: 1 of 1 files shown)


jaimergp (approved):
I thought the brackets were optional in `if` clauses but I can't find the docs on that now 🤷  Anyway, LGTM. Won't hurt to try!

jaimergp:
Nvm, I don't think this is related. It never got built because Windows CI never finished due to the issue being discussed at #8 

See [here](https://github.com/conda-incubator/conda-libmamba-solver/runs/5187078487?check_suite_focus=true)

jezdez:
Aaah, that makes sense!

PR #11, opened 2022-02-17
optimize logging a bit

jaimergp:
Follow up to #8 

`verbosity=3` is just too much output from libsolv (1.5M lines for complex solves), which hurts performance. By default, we will only have `verbosity=2` (wayy less! around 100 lines), but this can be cranked up with `conda`'s verbosity (`-vvv`) if needed.

Example of a "complex" solve with 1M lines of output (if `-vvv` is passed):

```
conda create -n demo napari tensorflow pytorch cudatoolkit=10.2 --override-channels -c conda-forge -c defaults --dry-run --experimental-solver=libmamba
```

---

Note, I also removed `CapturedDescriptor`, which wasn't being used anymore (and caused hangs in previous builds).

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -43,7 +43,7 @@ from .mamba_utils import (
     mamba_version,
 )
 from .state import SolverInputState, SolverOutputState, IndexHelper
-from .utils import CapturedDescriptor, CaptureStreamToFile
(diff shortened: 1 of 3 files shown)


jezdez:
Yeah, this is a good move, we can always ask about additional runs with higher verbosity via the issue form work in https://github.com/conda/conda/issues/11232. 👍🏻

PR #13, opened 2022-02-25
install mamba and deps explicitly on our CI

jaimergp:
Do not rely on `conda/conda` Docker images or `test/requirements.txt` configuration. Handle the extra deps and info messages here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           fetch-depth: 0
           repository: conda/conda
-          ref: libmamba
+          ref: libmamba # change to `master` once upstream is merged
           path: conda
 
       - name: Python ${{ matrix.python-version }}
@@ -51,8 +51,9 @@ jobs:
           -e TEST_SPLITS
           -e TEST_GROUP
           -e CONDA_EXPERIMENTAL_SOLVER
-          ghcr.io/conda/conda-ci:libmamba-linux-python${{ matrix.python-version }}
-          bash -c "/opt/conda/bin/python -m pip install /opt/conda-libmamba-solver-src --no-deps -vvv &&
+          ghcr.io/conda/conda-ci:master-linux-python${{ matrix.python-version }}
+          bash -c "/opt/conda/condabin/conda install -p /opt/conda 'conda-forge::mamba>=0.19.0' conda-forge::libcurl &&
+                   /opt/conda/bin/python -m pip install /opt/conda-libmamba-solver-src --no-deps -vvv &&
                    source /opt/conda-src/dev/linux/bashrc.sh &&
                    /opt/conda/bin/python -m pytest /opt/conda-libmamba-solver-src -vv"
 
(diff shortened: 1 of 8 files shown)


jaimergp:
Those _integration, group 2_ failures seem to be due to the availability of Python 3.10 on `defaults`.  The tests that fail are similar to this one:

```python
    def test_install_update_deps_flag(self):
        with make_temp_env("flask=2.0.1", "jinja2=3.0.1") as prefix:
            assert package_is_installed(prefix, "python=3.9")  # <---- this is the failing check
            assert package_is_installed(prefix, "flask=2.0.1")
            assert package_is_installed(prefix, "jinja2=3.0.1")

            run_command(Commands.INSTALL, prefix, "flask", "--update-deps")
            assert package_is_installed(prefix, "python=3.9")
            assert package_is_installed(prefix, "flask>2.0.1")
            assert package_is_installed(prefix, "jinja2>3.0.1")
```

Since 3.10 is (now?) available, it is picked up by the solver as the latest available (instead of 3.9, as it was the case before). I guess the intent of this test is to make sure that even with `--update-deps` and friends, Python stays pinned to its `major.minor` version (2nd check in the 2nd block). So we should maybe change that test to only annotate the Python version installed first, and _then_ check it stays the same after `--update-deps`. This test was also recently changed [in this PR](https://github.com/conda/conda/pull/11236).

cc @jezdez

jezdez:
> Those _integration, group 2_ failures seem to be due to the availability of Python 3.10 on `defaults`. The tests that fail are similar to this one:
> 
> ```python
>     def test_install_update_deps_flag(self):
>         with make_temp_env("flask=2.0.1", "jinja2=3.0.1") as prefix:
>             assert package_is_installed(prefix, "python=3.9")  # <---- this is the failing check
>             assert package_is_installed(prefix, "flask=2.0.1")
>             assert package_is_installed(prefix, "jinja2=3.0.1")
> 
>             run_command(Commands.INSTALL, prefix, "flask", "--update-deps")
>             assert package_is_installed(prefix, "python=3.9")
>             assert package_is_installed(prefix, "flask>2.0.1")
>             assert package_is_installed(prefix, "jinja2>3.0.1")
> ```
> 
> Since 3.10 is (now?) available, it is picked up by the solver as the latest available (instead of 3.9, as it was the case before). I guess the intent of this test is to make sure that even with `--update-deps` and friends, Python stays pinned to its `major.minor` version (2nd check in the 2nd block). So we should maybe change that test to only annotate the Python version installed first, and _then_ check it stays the same after `--update-deps`.

Yeah, that seems reasonable, I'll work on a patch upstream to fix that.

This test was also recently changed [in this PR](https://github.com/conda/conda/pull/11236).
> 
> cc @jezdez

jezdez:
Added the change in https://github.com/conda/conda/pull/11193/commits/631550eb8a90d08c2ac62c8e7c318c2db4b4366e.

jezdez:
I've merged https://github.com/conda/conda/pull/11193 and updated this PR.

jezdez (approved):
Thanks @jaimergp 🥳

jaimergp:
Ok, a new libmamba version was published with slightly different output: loggers now infix the library in the output. For example, before it was `info Parsing MatchSpec` and now it is `info libmamba Parsing MatchSpec`, hence the failures we were seeing.

Tests were adjusted and mamba requirements are now bumped to `>=0.22`.

jezdez (approved):
🚀

PR #14, opened 2022-03-01
Update label.

jezdez:


jaimergp:
This is weird, I don't think the branch is fully up-to-date. Can you rebase against `main`? This [line](https://github.com/conda-incubator/conda-libmamba-solver/blob/main/pyproject.toml#L90) is absent, for example.

PR #17, opened 2022-03-02
Ask for user confirmation before using libmamba solver

kenodegard:
Part 2/2 resolving https://github.com/conda/conda/issues/11224

Utilizes the `experimental_solver_confirmation` context value (see https://github.com/conda/conda/pull/11282) to determine whether the user has previously acknowledged the risks of running the experimental solver.

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -343,12 +343,20 @@ class LibMambaSolver(Solver):
                 confirm_yn(f"{msg}\n{risk}", dry_run=False)
 
                 # update ~/.condarc to no longer prompt
-                with open(user_rc_path, "r+") as fh:
-                    condarc = yaml_round_trip_load(fh.read())
-                    condarc["experimental_solver_confirmation"] = False
-                    fh.seek(0)
-                    fh.write(yaml_round_trip_dump(condarc))
-                reset_context()
+                try:
+                    with open(user_rc_path, "r+") as fh:
+                        condarc = yaml_round_trip_load(fh.read())
+                        condarc["experimental_solver_confirmation"] = False
+                        fh.seek(0)
+                        fh.write(yaml_round_trip_dump(condarc))
+                except OSError as e:
+                    log.warning(
+                        "Failed to set `experimental_solver_confirmation: false` in "
+                        "~/.condarc, is the home directory read-only?"
+                    )
+                    log.debug(e)
+                else:
+                    reset_context()
             else:
                 print(msg)
 
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -332,7 +332,7 @@ class LibMambaSolver(Solver):
                 ---------------------------------------------------------------
                 """
             ).rstrip()
-            if context.experimental_solver_confirmation:
+            if context.experimental_solver_confirmation and not context.dry_run:
                 risk = dedent(
                     """
                     There are risks using this experimental solver since it
@@ -344,7 +344,7 @@ class LibMambaSolver(Solver):
                     Do you wish to proceed
                     """
                 ).rstrip()
-                confirm_yn(f"{msg}\n{risk}", dry_run=False)
+                confirm_yn(f"{msg}\n{risk}")
 
                 # update ~/.condarc to no longer prompt
                 try:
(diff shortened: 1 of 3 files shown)


conda_libmamba_solver/solver.py
```diff
@@ -314,21 +315,42 @@ def _solving_loop(
...
+                """
+            ).rstrip()
+            if context.experimental_solver_confirmation:
+                risk = dedent(
+                    """
+                    There are risks using this experimental solver since it
+                    does not always produce identical results as that of the
+                    classical solver. As with any experimental/beta product it
+                    is not recommended to use this in production/critical
+                    environments.
```
jezdez, line 338:
```suggestion
                    There are risks using this experimental solver since it
                    does not always produce identical results as that of the
                    classical solver. As with any experiment it is *not*
                    recommended to use it in production or critical
                    environments.
```

We've tried to not use the "beta" terminology since it's perceived differently by users, really hard to understand what it means, while experiment is intentional about it being an unfinished product.

conda_libmamba_solver/solver.py
```diff
@@ -314,21 +315,42 @@ def _solving_loop(
...
-            )
+                ).rstrip()
+                confirm_yn(f"{msg}\n{risk}", dry_run=False)
+
+                # update ~/.condarc to no longer prompt
+                with open(user_rc_path, "r+") as fh:
+                    condarc = yaml_round_trip_load(fh.read())
+                    condarc["experimental_solver_confirmation"] = False
+                    fh.seek(0)
+                    fh.write(yaml_round_trip_dump(condarc))
```
jezdez, line 350: Should we wrap this in a try/except block with `OSError` to handle any accidental permission issues?

jezdez (approved):
Just a few notes around wording

kenodegard:
@jezdez @jaimergp when running something like:

```shell
python -m conda create --experimental-solver=libmamba -n m1 -c conda-forge vaex --dry-run
```

The experimental blurb is printed twice, do we know where the second print is occurring? And can this be changed to only print once?

This double print seems related to the `CondaSystemExit` exception (thrown when responding "no" to `confirm_yn`) being caught and results in the prompt occurring twice if the user says "no".

jezdez:
> @jezdez @jaimergp when running something like:
> 
> ```shell
> python -m conda create --experimental-solver=libmamba -n m1 -c conda-forge vaex --dry-run
> ```
> 
> The experimental blurb is printed twice, do we know where the second print is occurring? And can this be changed to only print once?
> 
> This double print seems related to the `CondaSystemExit` exception (thrown when responding "no" to `confirm_yn`) being caught and results in the prompt occurring twice if the user says "no".

Hm, sounds like the code needs to move in a different spot so that doesn't happen.

kenodegard:
Oh, I see, we are iterating over the `repodata_fns` (e.g. `('current_repodata.json', 'repodata.json')`) and creating a new solver for each one. In order to leave this experimental blurb here in this plugin it looks like we need to employ a global counter and only display the message on the first call.

This is also the same place where `CondaSystemExit` is getting caught:
https://github.com/conda/conda/blob/1905ece79f6d7d005911bf11a639503b1b2a215b/conda/cli/install.py#L287
Not sure yet how to allow the `CondaSystemExit` to propagate properly.

conda_libmamba_solver/solver.py
```diff
@@ -47,6 +48,7 @@
 from .utils import CaptureStreamToFile
 
 log = logging.getLogger(f"conda.{__name__}")
+BLURB_COUNT = 0
```
jaimergp, line 51:
```suggestion
_BLURB_COUNT = False
```

jaimergp:
> Oh, I see, we are iterating over the repodata_fns (e.g. ('current_repodata.json', 'repodata.json')) and creating a new solver for each one

We respect the outer loop but internally we are only using `repodata.json`, iirc. We shouldn't need to retry (if we did, it will fail again). This should noted as an issue so we can fix it!

conda_libmamba_solver/solver.py
kenodegard: I played with both and have no preference either way 🤷‍♂️

jaimergp: Let's leave it like it is then. No biggie.

jezdez (approved):
Looks good.

PR #18, opened 2022-03-08
Revert "Ask for user confirmation before using libmamba solver"

jaimergp:
Reverts conda-incubator/conda-libmamba-solver#17. See https://github.com/conda/conda/issues/11303

While this doesn't happen with `conda create`, it does happen with `conda install`, maybe other commands where existing envs are expected.


jezdez (approved):
Thank you!

PR #23, opened 2022-03-17
fix auth'd channels

jaimergp:
Closes https://github.com/conda/conda/issues/11344

We should probably refactor `testserver.sh` into something pytest-y for Windows compat, and possibly contribute that back to `mamba-org`.


wolfv:
We've used `pytest-xprocess` in the `powerloader` to start servers for testing and that works quite OK.

https://github.com/mamba-org/powerloader/blob/effe2b7e1f555616e4e4c877648658d1e6c89ded/test/helpers.py#L11

jaimergp:
There's a weird interaction between the new `test_channels` and `test_experimental::test_protection_for_base_env` for `libmamba`. It passes in isolation but not if test_channels had run, possibly due to side-effects of `run_command`?

jaimergp:
Thanks for the hint @wolfv 

@jezdez , this is ready for review.

@ myself: look into those weird `run_command` side-effects at some point. Not critical though, since it only affected the base env protections, which are only temporary.

jezdez (approved):
Love the fix/test ratio!

PR #26, opened 2022-03-21
Make sure noarch packages get reinstalled if python version changes 

jaimergp:
Closes https://github.com/conda/conda/issues/11346

---

- [X] Reproduce error with test. Note that upstream has a test for this, but defaults has a different pip (non-noarch), while conda-forge is noarch. This results in the same package.
- [x] Fix issue.


wolfv:
@jaimergp we've recently fixed that on micromamba side as well by passing additional package info data. 

https://github.com/mamba-org/mamba/blob/4b663ada019715ad4c989a1a57e81fe12edde3d8/mamba/mamba/mamba.py#L207-L215

there will be an API change for the next mamba version though.

jezdez (approved):
This is surprising but good we caught it!

PR #32, opened 2022-04-25
fix escape_channel_url problems

jaimergp:
A potential solution for the Windows issues we are seeing. This brings `conda.common.url.escape_channel_url` back to our codebase instead of letting it live on `conda/conda`, where updating it might be trickier. Same code as https://github.com/conda/conda/pull/11416


jezdez (approved):

jezdez (approved):
If we have some tests to transfer over that'd be useful to have.

PR #33, opened 2022-04-25
Faster Python version changes

jaimergp:
Right now, we approach frozen installs like this:

* By default, `conda` tells us to freeze everything
* We try to solve, probably failing
* Parse the exception to know what failed
* Annotate the failure as a conflict, which will be special cases during the solver preparation: conflicting specs are relaxed so the solver can choose different builds / version

This has a problem when a LOT of packages conflict, because we need to try once for each installed package. An example where this happens often is a Python version change: all the installed packages that depend on the old version are potential conflicts! 

Since this is relatively common scenario, this PRs special cases this particular situation and annotates all immediate Python-depending packages (those that explicitly list `python` or `python_abi` in their dependencies) as conflicts (before it happens) when the installed Python cannot be satisfied by the incoming request.

Some rough timings show the difference for the `base` environment of the Docker image used in CI (169 packages):

Before this PR:

```
real    1m46.725s
user    1m45.140s
sys     0m1.814s
```

After this PR:

```
real    0m13.520s
user    0m12.560s
sys     0m0.347s
```

You'll note I also added a special `CONDA_LIBMAMBA_MAX_ATTEMPTS` environment variable to control the number of solving attempts we try for freeze_installed before giving up. It defaults to the number of installed packages to ensure best compatibility with conda's behaviour, but in some cases the user might not care for the sake of performance. This env var is not parsed by the `Context` object - we just take it as is, if present.

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -255,7 +255,10 @@ class LibMambaSolver(Solver):
     def _solving_loop(
         self, in_state: SolverInputState, out_state: SolverOutputState, index: LibMambaIndexHelper,
     ):
-        for attempt in range(1, max(1, len(in_state.installed)) + 1):
+        max_attempts = max(
+            2, int(os.environ.get("CONDA_LIBMAMBA_MAX_ATTEMPTS", len(in_state.installed))) + 1,
+        )
+        for attempt in range(1, max_attempts):
             log.debug("Starting solver attempt %s", attempt)
             if not context.json and not context.quiet and os.environ.get("EXTRA_DEBUG_TO_STDOUT"):
                 print("----- Starting solver attempt", attempt, "------", file=sys.stderr)
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -452,8 +452,8 @@ class LibMambaSolver(Solver):
         # ## When the Python version changes, this implies all packages depending on
         # ## python will be reinstalled too. This can mean that we'll have to try for every
         # ## installed package to result in a conflict before we get to actually solve everything
-        # ## A workaround is to let all python-depending specs to "float" (no build constrain) if
-        # ## the python version was found to change
+        # ## A workaround is to let all non-noarch python-depending specs to "float" by marking
+        # ## them as a conflict preemptively
         python_version_might_change = False
         installed_python = in_state.installed.get("python")
         to_be_installed_python = out_state.specs.get("python")
@@ -469,8 +469,11 @@ class LibMambaSolver(Solver):
             installed = in_state.installed.get(name)
             key = "INSTALL", api.SOLVER_INSTALL
 
-            ## Fast-track python version changes
+            # Fast-track Python version changes: mark non-noarch Python-depending packages as
+            #  conflicting (see `python_version_might_change` definition above for more details)
             if python_version_might_change and installed is not None:
+                if installed.noarch is not None:
+                    continue
                 for dep in installed.depends:
                     dep_spec = MatchSpec(dep)
                     if dep_spec.name in ("python", "python_abi"):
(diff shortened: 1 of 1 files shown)


conda_libmamba_solver/solver.py
```diff
@@ -255,7 +255,10 @@ def solve_final_state(
     def _solving_loop(
         self, in_state: SolverInputState, out_state: SolverOutputState, index: LibMambaIndexHelper,
     ):
-        for attempt in range(1, max(1, len(in_state.installed)) + 1):
+        max_attempts = max(
+            2, int(os.environ.get("CONDA_LIBMAMBA_MAX_ATTEMPTS", len(in_state.installed))) + 1,
```
jezdez, line 259:
```suggestion
            2, int(os.environ.get("CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS", len(in_state.installed))) + 1,
```

conda_libmamba_solver/solver.py
```diff
@@ -444,23 +447,54 @@ def _specs_to_tasks_add(self, in_state: SolverInputState, out_state: SolverOutpu
...
+        # Fast-track python version changes
+        # ## When the Python version changes, this implies all packages depending on
+        # ## python will be reinstalled too. This can mean that we'll have to try for every
+        # ## installed package to result in a conflict before we get to actually solve everything
+        # ## A workaround is to let all non-noarch python-depending specs to "float" by marking
+        # ## them as a conflict preemptively
+        python_version_might_change = False
+        installed_python = in_state.installed.get("python")
+        to_be_installed_python = out_state.specs.get("python")
+        if installed_python is not None and to_be_installed_python is not None:
```
jezdez, line 460:
```suggestion
        if all((installed_python, to_be_installed_python)):
```

conda_libmamba_solver/solver.py
```diff
@@ -444,23 +447,54 @@ def _specs_to_tasks_add(self, in_state: SolverInputState, out_state: SolverOutpu
...
         for name, spec in out_state.specs.items():
             if name.startswith("__"):
                 continue
             self._check_spec_compat(spec)
             spec_str = str(spec)
+            installed = in_state.installed.get(name)
             key = "INSTALL", api.SOLVER_INSTALL
+
+            # Fast-track Python version changes: mark non-noarch Python-depending packages as
+            #  conflicting (see `python_version_might_change` definition above for more details)
```
jezdez, line 473:
```suggestion
            # conflicting (see `python_version_might_change` definition above for more details)
```

jezdez: @chenghlee Hey, do you know if this change will actually be moot by our plans to deprecate of noarch for interpreter based packages?

conda_libmamba_solver/solver.py
```diff
@@ -444,23 +447,54 @@ def _specs_to_tasks_add(self, in_state: SolverInputState, out_state: SolverOutpu
...
-            if name in in_state.installed:
-                installed = in_state.installed[name]
+
+            if installed is not None:
                 # ## Regular task ###
                 key = "UPDATE", api.SOLVER_UPDATE
+
                 # ## Protect if installed AND history
-                if name in protected:
+                if name in protected and name:
```
jezdez, line 497:
```suggestion
                if name and name in protected:
```

jezdez (commented):
I think this makes sense, but had a few suggestions..

jaimergp:
Hm, I think I need to special case which specs get their build constrains dropped. I think if the user tries to do something like `conda install numpy=*=*py38*`, it could be dropped to `numpy`, resulting in a different solve (e.g. something with Python 3.9 instead of 3.8). This will need some tests to ensure this behaviour is ok.

conda_libmamba_solver/solver.py
jaimergp: There are plans to deprecate the `noarch: python` functionality? I wasn't aware of this. Is there a CEP covering the rationale? It seems like a pretty big change.

jaimergp:
> Hm, I think I need to special case which specs get their build constrains dropped. I think if the user tries to do something like `conda install numpy=*=*py38*`, it could be dropped to `numpy`, resulting in a different solve (e.g. something with Python 3.9 instead of 3.8). This will need some tests to ensure this behaviour is ok.

Ok, I've run some tests locally and this has nothing to do with this PR. I'll file a separate issue.

jezdez (approved):

PR #35, opened 2022-04-27
libmamba 0.23 compatibility

jaimergp:
libmamba 0.23 introduces API breaking changes. This PR adds the necessary modifications to ensure our integrations keep working!

PR #36, opened 2022-05-02
Fix handling of `*`-enabled build strings

jaimergp:
Closes https://github.com/conda/conda/issues/11450

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -468,8 +468,16 @@ class LibMambaSolver(Solver):
 
     @staticmethod
     def _spec_to_str(spec):
+        """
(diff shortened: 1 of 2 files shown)


jaimergp:
Looks like this is one more instance of https://github.com/conda/conda/pull/9980 and the lossy behaviour of `str` and `MatchSpec`.

This is what I am seeing. We switched from `spec.conda_build_form()` to `str(spec)` to fix the "package from channel X" spec situation. If we switch back to `.conda_build_form()`, this issue is "fixed" (but we regress with the channel problem).

Since we have a similar problem (post-processing the str'd MatchSpec) on #35, I'll wait til that one is fixed to submit the adequate post-processing here... but yep, those issues with MatchSpec round trips should really be fixed!

PR #38, opened 2022-05-05
add black config to pyproject.toml

jaimergp:
Turns out I had my IDE configured to _always_ use `black -l 99` _if_ no config was present on the repo, so we have been effectively using `line-length = 99` since the inception of the repo.

I guess it doesn't hurt to keep using that setting, officially.

I guess (bis) adding some CI to ensure formatting is correct would help as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,9 @@ build-backend = "flit_core.buildapi"
 [tool.flit.module]
 name = "conda_libmamba_solver"
 
+[tool.black]
+line-length = 99
+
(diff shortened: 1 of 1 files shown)


jezdez (approved):

jezdez:
BTW in the conda org we've been standardizing around precommit and [precommit CI](pre-commit.ci) since it makes running things locally easier and doesn't only work with Python. In case you want to spend more time on this, I think it would be worth adopting that here, too.

PR #42, opened 2022-06-03
add tests for user agent strings

jaimergp:
This was fun :)

With conda 4.13 and libmamba 0.23 we can customize the user agent to assess the amount of usage the experimental solver is getting out there. I have added specific tests for all possible combinations, as seen from the client and the server. Check the docstrings for more info :)


tests/test_user_agent.py
```diff
@@ -0,0 +1,237 @@
...
+    error = "RuntimeError: Multi-download failed. Reason: Transfer finalized, status: 404"
+    for line in stdout.splitlines():
+        if line.strip().startswith(error):
+            _, address_and_suffix = line.split("[")
+            address, _ = address_and_suffix.split("]")
+            b64_headers = address.split("/")[-1]
+            headers = base64.b64decode(b64_headers).decode("utf-8")
+            break
+    else:
+        raise RuntimeError("Couldn't find the error message containing payload")
```
jezdez, line 109:
```suggestion
        raise AssertionError("Couldn't find the error message containing payload")
```

tests/test_user_agent.py
```diff
@@ -0,0 +1,237 @@
...
+    for line in headers.splitlines():
+        if line.startswith("User-Agent:"):
+            assert "conda/" in line
+            assert "solver/libmamba" in line
+            assert f"conda-libmamba-solver/{version('conda-libmamba-solver')}" in line
+            assert f"libmambapy/{version('libmambapy')}" in line
+            assert "libcurl/" in line
+            break
+    else:
+        raise RuntimeError("Couldn't find the User-Agent info in headers!")
```
jezdez, line 120:
```suggestion
        raise AssertionError("Couldn't find the User-Agent info in headers!")
```

tests/test_user_agent.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def check_user_agent_requests(stdout, solver, request_type="repodata"):
```
jezdez, line 123:
```suggestion
def assert_requests_user_agent(stdout, solver, request_type="repodata"):
```

tests/test_user_agent.py
```diff
@@ -0,0 +1,237 @@
...
+            stdout=PIPE,
+            stderr=PIPE,
+            text=True,
+            env=env,
+        )
+    print("-- STDOUT --")
+    print(process.stdout)
+    print("-- STDERR --")
+    print(process.stderr)
+    check_user_agent_requests(process.stdout, solver="classic", request_type="packages")
```
jezdez, line 237:
```suggestion
def test_user_agent_classic_packages(server_auth_none_debug_packages, tmp_path):
    run([sys.executable, "-m", "conda", "clean", "--tarballs", "--packages", "--yes"])

    env = os.environ.copy()
    env["CONDA_PKGS_DIRS"] = str(tmp_path)
    process = create_with_channel(
        server_auth_none_debug_packages,
        solver="classic",
        check=False,
        stdout=PIPE,
        stderr=PIPE,
        text=True,
        env=env,
    )
    print("-- STDOUT --")
    print(process.stdout)
    print("-- STDERR --")
    print(process.stderr)
    assert_requests_user_agent(process.stdout, solver="classic", request_type="packages")

```

tests/test_user_agent.py
```diff
@@ -0,0 +1,237 @@
...
+    # the output might contain more than one json dict
+    # tip: conda separates json payloads with nul (\x00)
+    # we split by nul and keep the last chunk surrounded by {}
+    payload = None
+    for chunk in stdout.split("\x00"):
+        chunk = chunk.strip()
+        if chunk.startswith("{") and chunk.endswith("}"):
+            payload = chunk
+    if payload is None:
+        raise ValueError(f"Could not find a JSON payload in the output: {stdout}")
```
jezdez, line 139:
```suggestion
        raise AssertionError(f"Could not find a JSON payload in the output: {stdout}")
```

dev/requirements.txt
```diff
@@ -1,2 +1,3 @@
-libmambapy>=0.23
+conda-forge::libmambapy>=0.23
 conda-forge::libcurl
+conda-forge::libarchive
```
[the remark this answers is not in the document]
jaimergp: I tried but it won't solve :(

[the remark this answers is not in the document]
jaimergp: It does solve but we have the libarchive SONAME incompatibility:

https://github.com/conda-incubator/conda-libmamba-solver/runs/6797972107?check_suite_focus=true#step:4:351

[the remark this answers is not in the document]
jaimergp: It seems to work now if we don't force libcurl from conda-forge (we had to before because it was the only build with unicode on windows, but defaults has that too now!). So I'll merge once everything is passing

PR #45, opened 2022-07-08
Deprecate libmamba-draft

jaimergp:
This implementation was never announced publicly (e.g. tweet, blog post, etc) and we only kept it around for debugging purposes. In practice we didn't have to use it ever but it was maintained regardless.

I think we can deprecate it now. If there are any users of this instance out there, they will be forwarded to the new solver (which should be equivalent in functionality, or better) with a little warning.

This also allows us to remove some utility functions in `mamba_utils` that were only there to support this implementation.


jezdez (approved):
This looks good in general, just need to look into the test failures.

PR #46, opened 2022-07-08
Improve post-solve performance by reimplementing "conda is outdated" notifications

jaimergp:
conda calls `_notify_conda_outdated` _after_ the environment is solved to see if we could have updated conda in that step. Since conda-libmamba-solver only reimplemented `solve_final_state` and that function is called in `solve_for_diff`, we were blindly using it.

It shouldn't have been a problem if this function didn't involve loading all the channels again. In conda classic it's not a problem because `SubdirData` classes cache the channels during the process life, but since `conda-libmamba-solver` loads channels with libmamba, we didn't have that cache and had to load everything again, resulting in 20s of overhead.

What we have done is to reimplement this function so it uses the libmamba-loaded channels. These are only available in the `solve_final_state` method, so we have added a couple of arguments and checks to detect which function is calling it. Not perfect, but simpler than other options. Ideally `conda` classic would define a hook to subclass and override too, but we are not there yet.

Before:

image

After:

image

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -199,7 +199,7 @@ class LibMambaSolver(Solver):
             out_state = self._solving_loop(in_state, out_state, index)
 
         # Restore intended verbosity to avoid unwanted
-        # "freeing xxxx..." messages when the libmambpy objects are deleted
+        # "freeing xxxx..." messages when the libmambapy objects are deleted
         api_ctx.verbosity = context.verbosity
         api_ctx.set_verbosity(context.verbosity)
 
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -782,6 +782,7 @@ class LibMambaSolver(Solver):
         if not context.notify_outdated_conda or context.quiet:
             # This check can be silenced with a specific option in the context or in quiet mode
             return
+
         current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
         if not current_conda_prefix_rec:
             # We are checking whether conda can be found in the environment conda is
@@ -791,9 +792,15 @@ class LibMambaSolver(Solver):
         channel_name = current_conda_prefix_rec.channel.canonical_name
         if channel_name == UNKNOWN_CHANNEL:
             channel_name = "defaults"
-        # only look for a newer conda in the channel conda is currently installed from
-        # conda_newer_str = f"{channel_name}::conda>{current_conda_prefix_rec.version}"
-        conda_newer_str = f"{channel_name}::conda>4.10"
+
+        # only check the loaded index if it contains the channel conda should come from
+        # otherwise ignore
+        index_channels = set(getattr(chn, "canonical_name", chn) for chn in index._channels)
+        if channel_name in index_channels:
+            return
+
+        # we only want to check if a newer conda is available in the channel we installed it from
+        conda_newer_str = f"{channel_name}::conda>{current_conda_prefix_rec.version}"
         conda_newer_spec = MatchSpec(conda_newer_str)
 
         # if target prefix is the same conda is running from
@@ -803,7 +810,7 @@ class LibMambaSolver(Solver):
             if any(conda_newer_spec.match(record) for record in final_state):
                 return
 
-        # check if the index contains records that match a more recent conda version
+        # check if the loaded index contains records that match a more recent conda version
         conda_newer_records = index.search(conda_newer_str)
 
         # print instructions to stderr if we found a newer conda
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -8,17 +8,25 @@ from itertools import chain
 from collections import defaultdict
 import logging
 import sys
-from typing import Mapping, Optional
+from typing import Mapping, Optional, Iterable
 from textwrap import dedent
 import re
 from functools import lru_cache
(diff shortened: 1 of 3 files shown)


conda_libmamba_solver/solver.py
```diff
@@ -157,27 +165,44 @@ def _channel_to_url_or_name(channel):
...
-            for pkg in result["result"]["pkgs"]:
-                explicit_pool.add(pkg["name"])
+            pkg_records = self.depends(spec.dist_str())
+            for record in pkg_records:
+                explicit_pool.add(record.name)
         return tuple(explicit_pool)
 
+    def _process_query_result(self, result_str, records=True):
+        result = json_load(result_str)
+        if result["result"]["status"] != "OK":
```
jezdez, line 193: Should this be written more defensively to catch malformed JSON?

conda_libmamba_solver/solver.py
```diff
@@ -239,6 +264,68 @@ def __init__(
...
+            return
+        current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
+        if not current_conda_prefix_rec:
+            return
+
+        channel_name = current_conda_prefix_rec.channel.canonical_name
+        if channel_name == UNKNOWN_CHANNEL:
+            channel_name = "defaults"
+        # only look for a newer conda in the channel conda is currently installed from
+        conda_newer_str = f"{channel_name}::conda>4.10"
```
jezdez, line 292: Could you elaborate why >4.10 is used here?

conda_libmamba_solver/solver.py
```diff
@@ -239,6 +264,68 @@ def __init__(
...
+        That way we can reuse the SolverInputState instance we have lying around,
+        which contains the index information we need, before we lose it.
+        """
+        if index is None and final_state is None:
+            return
+        if not context.notify_outdated_conda or context.quiet:
+            return
+        current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
+        if not current_conda_prefix_rec:
+            return
```
jezdez, line 286: Could you add comments on when these conditions are met?

conda_libmamba_solver/solver.py
```diff
@@ -239,6 +264,68 @@ def __init__(
...
+
+                    Please update conda by running
+
+                        $ conda update -n base {add_channel}conda
+
+                    """
+                ),
+                file=sys.stderr,
+            )
+
```
jezdez, line 328: With the new release policies and improved release times, I wonder if it wouldn't be useful to also print the channel name for non-defaults channels here?

```suggestion
            print(
                dedent(
                    f"""

                    ==> WARNING: A newer version of conda exists. <==
                        current version: {_conda_version}
                        latest version: {latest_version}

                    Please update conda by running

                        $ conda update -n base -c {channel_name} conda

                    """
                ),
                file=sys.stderr,
            )

```

jezdez (approved):
This looks good in general, just minor nits, thank you @jaimergp!

conda_libmamba_solver/solver.py
jaimergp: Oops, I think it was me debugging stuff. It should use the currently installed version.

conda_libmamba_solver/solver.py
jaimergp: Let's see how this looks :)

jaimergp:
Errors due to https://github.com/conda/conda/pull/11670 upstream.

conda_libmamba_solver/solver.py
jaimergp: Nice! Btw, this will need to be updated in [`conda`'s implementation too](https://github.com/conda/conda/blob/e69b9efd5fc96e26bb2aa3b180e3c504c65ce65f/conda/core/solve.py#L1019-L1029), @jezdez

jaimergp:
Upstream errors now due to the S3 tests we added, which will be handled by #49

conda_libmamba_solver/solver.py
```diff
@@ -738,3 +754,81 @@ def _called_from_conda_build(self):
...
+        # only check the loaded index if it contains the channel conda should come from
+        # otherwise ignore
+        index_channels = set(getattr(chn, "canonical_name", chn) for chn in index._channels)
+        if channel_name in index_channels:
+            return
+
+        # we only want to check if a newer conda is available in the channel we installed it from
+        conda_newer_str = f"{channel_name}::conda>{current_conda_prefix_rec.version}"
+        conda_newer_spec = MatchSpec(conda_newer_str)
```
jaimergp, line 804: @jezdez I added this extra bit now. I hope that's ok!

conda_libmamba_solver/solver.py
```diff
@@ -738,3 +754,81 @@ def _called_from_conda_build(self):
...
+
+        channel_name = current_conda_prefix_rec.channel.canonical_name
+        if channel_name == UNKNOWN_CHANNEL:
+            channel_name = "defaults"
+
+        # only check the loaded index if it contains the channel conda should come from
+        # otherwise ignore
+        index_channels = set(getattr(chn, "canonical_name", chn) for chn in index._channels)
+        if channel_name in index_channels:
+            return
```
jaimergp, line 800: @jezdez I added this extra bit now. I hope that's ok!

jezdez: Yeah, that seems sensible.

jezdez (approved):

PR #48, opened 2022-07-19
Remove debugging code

jaimergp:
Cleaning up:

* some old blocks we never used after release (they helped debugging by printing logs directly to stderr)
* captured stdout from libmamba (the content that went into the log file)
* updated the disclaimer at the top to not mention the log file

We also have to remove the references to the log file from the `conda.base.context` object (where the path to the hypothetical log file was generated) and the issue template in `conda/conda`.


jezdez (approved):
I love these PRs! <3

PR #49, opened 2022-07-19
Refactor index helpers

jaimergp:
This PR adds a fallback behaviour for conda channels the libmamba repodata fetchers don't support yet (e.g. `s3://`). It would be trivial to adapt this other rules if needed (e.g. on-prem stuff, special auth).

If the channel contains a scheme and this scheme is not http[s] or file, we will delegate to conda's `SubdirData`. To ensure priorities are respected, we fetch channels in order. 

Side note: the adapter code used would allow to totally get by without libmamba loaders if needed, too.

Note conda doesn't have tests where s3 channels are actually fetched.


tests/test_channels.py
```diff
@@ -1,32 +1,157 @@
...
+        return resp
+
+    return _send_boto3
+
+
+@pytest.skipif(
+    run(["minio", "-v"], check=False).returncode != 0,
+    reason="Minio server not available on PATH",
+)
+def test_s3_server(s3_server):
```
jaimergp, line 144: @jezdez I am wondering whether this mocked channel tests would be useful upstream too. Most tests just use anaconda.org and that's it. We don't have S3 or FTP specific tests in `conda/conda`.

jezdez: Yes, that'd be splendid!

jaimergp:
I am [upstreaming the S3 tests to conda/conda](https://github.com/conda/conda/pull/11644). Once that one is merged, we can remove it here, so let's not merge this one yet!

jaimergp:
Hm, unclear what's wrong now, since the test is essentially the same?

For some reason the download is expected to be empty, but then it's not, but then some files are reported missing and things break 🤷 

I'll take a look tomorrow, with a fresher mind.

.github/workflows/upstream_ci.yml
```diff
@@ -127,25 +127,18 @@ jobs:
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
 
       - name: Setup environment
-        shell: bash -l {0}
+        shell: bash -el {0}
```
jezdez, line 130: This makes it fail if there is no `dev/macos/setup.sh`?

tests/channel_testing_utils.py
```diff
@@ -1,22 +1,23 @@
...
 
 import pytest
 from xprocess import ProcessStarter
 
-from conda.testing.integration import _get_temp_prefix
+from conda.testing.integration import _get_temp_prefix, run_command
 
 
-def _mock_server(xprocess, name, port, auth="none", user=None, password=None, token=None):
+def _mock_http_server(xprocess, name, port, auth="none", user=None, password=None, token=None):
```
jezdez, line 14: Just tiny naming thing, since this is not related to the `mock` library I'd rather not use it for this fixture to reduce confusion when we use it in the future.

jezdez (approved):
Looks good to me, just minor questions.

.github/workflows/upstream_ci.yml
jaimergp: Yes. This is an oversight we had at `setup-miniconda` and I am slowly adopting the new `shell` across projects. [See issue here](https://github.com/conda-incubator/setup-miniconda/issues/227).

PR #52, opened 2022-09-13
Load repodata from historically used channels (as stored in `PrefixData`)

jaimergp:
Closes https://github.com/conda/conda/issues/11790

--- 

I am not sure whether we should go this way, though. I think it's a good idea, but we also need to enable this on `conda/conda`, `mamba` and `micromamba` for better compatibility across the ecosystem.

Authenticated channels are the only blocker here, but it can be either worked around or handled gracefully with good errors.

### DIFF
--- a/conda_libmamba_solver/index.py
+++ b/conda_libmamba_solver/index.py
@@ -51,11 +51,14 @@ class LibMambaIndexHelper(IndexHelper):
         # See https://github.com/conda/conda/issues/11790
         channels_from_installed = []
         for record in installed_records:
-            if record.channel.auth or record.channel.token or record.channel.name == "@":
+            if record.channel.auth or record.channel.token:
                 # skip if the channel has authentication info, because
                 # it might cause issues with expired tokens and what not
-                # if the name is @, that means is a virtual package -- skip it too
                 continue
+            if record.channel.name in ("@", "<develop>", "pypi"):
+               # These "channels" are not really channels, more like
+               # metadata placeholders
+               continue
             if record.channel.subdir_url not in channels_from_installed:
                 channels_from_installed.append(record.channel.subdir_url)
         if channels_from_installed:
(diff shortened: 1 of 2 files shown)


tests/test_channels.py
```diff
@@ -47,3 +49,33 @@ def test_channel_matchspec():
...
+
+def test_channels_prefixdata():
+    """
+    Make sure libmamba does not complain about missing channels
+    used in previous commands.
+
+    See https://github.com/conda/conda/issues/11790
+    """
+    with make_temp_env("conda-forge::xz", "python") as prefix:
+        output = check_output(
```
jaimergp, line 62: Note we need a whole subprocess here because libmamba prints to STDERR directly, and that won't be intercepted with our naïve in-process capturing mechanisms.

conda_libmamba_solver/index.py
```diff
@@ -43,10 +43,27 @@ def __init__(
...
         self._repos.append(installed_repo)
 
+        # See https://github.com/conda/conda/issues/11790
+        channels_from_installed = []
+        for record in installed_records:
+            if record.channel.auth or record.channel.token:
+                # skip if the channel has authentication info, because
+                # it might cause issues with expired tokens and what not
+                continue
+            if record.channel.name in ("@", "<develop>", "pypi"):
```
jezdez, line 58: Do you know if we have a definite list of these pseudo channel names in conda that we don't have to hard-code it here?

jezdez (approved):
LGTM in general, just a minor note about hard-coding the pseudo channel names. Thank you @jaimergp!

conda_libmamba_solver/index.py
jaimergp: I checked but I couldn't find it. Looks like they were added as needed for edge cases found during the development of orthogonal features. `@` is for virtual channels, `<develop>` seems to come from `pip install -e` and `pypi` is found for `pip install` that do not have a `conda-meta` entry, as far as I can understand.

jezdez (approved):

jaimergp:
@jezdez Should we port this over to `conda/conda`?

PR #55, opened 2022-10-19
still experimental in current conda versions

dholth:
I was referencing this feature and noticed it was still `--experimental` in the version of conda I have installed.


jaimergp:
The new release (yet to be planned) will pin for conda 22.11+, but yes, adding this to the docs already is a good idea, thanks!

dholth:
I don't have write access, please merge when tests pass.

PR #62, opened 2022-11-10
Update CI test config

jezdez:
### Description

- adds pre-commit support to improve automation and create a more consistent dev/prod parity

- applies black and isort automatically and adds copyright disclaimers to files
- adds Python 3.10 testing and renames the CI workflow to tests
- add a test analyze step
- build canaries inline with the tests (see https://github.com/conda/conda/pull/12040)


tests/channel_testing_utils.py
```diff
@@ -1,14 +1,15 @@
-import pathlib
+# Copyright (C) 2022 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
```
jaimergp, line 2: This module has some functions adapted from Powerloader (same license, different copyright holder). I am not sure how this is handled here, but better mention it in case you know?

tests/_reposerver.py
```diff
@@ -1,3 +1,5 @@
+# Copyright (C) 2022 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
```
jaimergp, line 2: Same issue as before. This module vendors code from `mamba`.

jaimergp (approved):
Nice! This will be lovely to have.

I just added some comments about the copyright lines for some modules that vendor code from other parties. Once addressed, we can merge.

PR #78, opened 2022-11-25
Ensure compatibility with libmamba 1.0 and conda 22.11

### LINKED ISSUE #57: Update to use libmamba 1.0
mamba/libmamba 1.0 [was released](https://wolfv.medium.com/releasing-mamba-1-0-8e3c52cc6d37) and we need to update the package to make sure of that.

jaimergp:
Fixes #57.

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -757,7 +757,7 @@ class LibMambaSolver(Solver):
             return
 
         channel_name = current_conda_prefix_rec.channel.canonical_name
-        if channel_name in (UNKNOWN_CHANNEL, "@", "<develop>", "pypi"):
+        if channel_name in (UNKNOWN_CHANNEL, "<develop>", "pypi"):
             channel_name = "defaults"
 
         # only check the loaded index if it contains the channel conda should come from
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -757,13 +757,13 @@ class LibMambaSolver(Solver):
             return
 
         channel_name = current_conda_prefix_rec.channel.canonical_name
-        if channel_name == UNKNOWN_CHANNEL:
+        if channel_name in (UNKNOWN_CHANNEL, "@", "<develop>", "pypi"):
             channel_name = "defaults"
 
         # only check the loaded index if it contains the channel conda should come from
         # otherwise ignore
         index_channels = {getattr(chn, "canonical_name", chn) for chn in index._channels}
-        if channel_name in index_channels:
+        if channel_name not in index_channels:
             return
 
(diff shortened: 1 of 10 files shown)


jaimergp:
Python 3.7 tests will fail because `conda-forge` won't build for Python 3.7 anymore, and libmamba 1.0 is not available in that Python version. This will pass once we have libmamba 1.0 on `defaults` and we can switch back to that for packaging.

Otherwise we can also drop 3.7 from the matrix, but I think this is still supported by Anaconda so... let's wait :)

conda_libmamba_solver/solver.py
```diff
@@ -757,13 +757,13 @@ def _notify_conda_outdated(
...
         channel_name = current_conda_prefix_rec.channel.canonical_name
-        if channel_name == UNKNOWN_CHANNEL:
+        if channel_name in (UNKNOWN_CHANNEL, "@", "<develop>", "pypi"):
             channel_name = "defaults"
 
         # only check the loaded index if it contains the channel conda should come from
         # otherwise ignore
         index_channels = {getattr(chn, "canonical_name", chn) for chn in index._channels}
-        if channel_name in index_channels:
+        if channel_name not in index_channels:
```
jaimergp, line 766: Apparently we had this logic reversed 😨 

This now will indeed check where there's a conda update if the solution includes conda, and we happened to have the channel conda came from in the loaded channels already (to prevent additional fetching).

conda_libmamba_solver/solver.py
```diff
@@ -757,13 +757,13 @@ def _notify_conda_outdated(
             return
 
         channel_name = current_conda_prefix_rec.channel.canonical_name
-        if channel_name == UNKNOWN_CHANNEL:
+        if channel_name in (UNKNOWN_CHANNEL, "@", "<develop>", "pypi"):
```
jaimergp, line 760: Fixes a query issue when we are checking whether a conda update is available, but its source channel is "develop" because it was installed in dev mode. I added other commonly used fake channels just in case.

conda_libmamba_solver/solver.py
jezdez: Yikes, that seems bad, will take a closer look at what that means..

jezdez:
@jaimergp we're down to one error: `tests/test_channels.py::test_http_server_auth_basic_email` Not sure what's causing this, but it fails when the user has a @ in it?

conda_libmamba_solver/solver.py
jezdez: Will this @ here stumble over authentication passed in?

jezdez:
pre-commit.ci autofix

jezdez (approved):

PR #127, opened 2023-02-04
Change build backend from flit to hatchling

jaimergp:
### Checklist - did you ...

I forgot to update this attribute in the last release, so just fixing it now so it's clear in the future. Internally, `conda-libmamba-solver.__version__` will report `22.8.1` til this is released. To be clear, `conda list` does report the correct version. This now uses `hatchling` plus a setuptools-scm plugin, `hatch-vcs`, as the build backend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,6 @@ cython_debug/
 
 # rever
 rever/
+
+# setuptools_scm
+_version.py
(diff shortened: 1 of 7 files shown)


jezdez:
@jaimergp We could switch to setuptools_scm via https://pypi.org/project/flit-scm/ or (while looking at https://github.com/pypa/flit/issues/257) switch to PyPA's hatch instead that has setuptools_scm support?

jezdez (approved):

PR #136, opened 2023-02-20
Manual workflow_dispatch test

costrouc:
### Description

Adding a github-action where we can launch conda/conda-libmamba-solver on any:
  - [ ] conda version/branch
  - [X] conda-libmamba-solver version/branch
  - [X] any os linux/macos/windows

Opens up a tmate session that is reachable via ssh (see CI) with given versions installed on given os. Only the user who triggered the CI can access the terminal via ssh (using ssh keys). Normally I don't trust github-actions from users but [mxschmitt/action-tmate](https://github.com/mxschmitt/action-tmate) looks heavily depended upon and several well known people in the python space are using this.

image

### Checklist - did you ...

### DIFF
--- a/.github/workflows/manual-test.yaml
+++ b/.github/workflows/manual-test.yaml
@@ -1,0 +1,76 @@
+name: Manual Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      conda-version:
+        description: 'Conda Version'
+        required: true
+        default: '23.1.0'
+        type: choice
+        options:
+        - 'main'
+        - '23.1.0'
+      conda-libmamba-version:
+        description: 'conda-libmamba-solver version/branch'
+        required: true
+        default: '23.1.0'
+      channel:
+        description: 'Conda Channel'
+        required: true
+        default: 'conda-forge'
+      os:
+        description: 'Operating System'
+        required: true
+        default: 'ubuntu-latest'
+        type: choice
+        options:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+
+jobs:
+  conda-libmamba-solver:
+    name: conda-libmamba-solver=${{ inputs.conda-libmamba-version }} os=${{ inputs.os }}
+    runs-on: ${{ inputs.os }}
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@master
+
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-activate-base: true
+          activate-environment: ""
+          conda-version: ${{ inputs.conda-version }}
+
+      - name: conda-libmamba-solver from Conda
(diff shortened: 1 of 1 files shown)


costrouc:
@jaimergp would it be possible to install conda from source in this action as well? 

I'm guessing there would be a step to 

1. git clone conda/conda
2. git checkout <branch>
3. python -m conda init --dev bash

This follows roughly what `dev/linux/bashrc.sh` does for conda. Though I don't see the equivalent for conda in OSX and windows.

jaimergp:
Installing from source require three different methods (one per OS), as you can see in either `tests.yml` or `upstream_tests.yml`. You can borrow the steps from those ones if you want!

.github/workflows/manual-test.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-activate-base: true
+          activate-environment: ""
+          conda-version: ${{ inputs.conda-version }}
+
+      - name: conda-libmamba-solver from Conda
+        shell: bash -el {0}
+        run: conda install -c conda-forge conda-libmamba-solver==${{ inputs.conda-libmamba-version }}
+        if: ${{ startsWith(inputs.conda-libmamba-version, '2') }}
```
jaimergp, line 49:
```suggestion
        if: ${{ startsWith(inputs.conda-libmamba-version, '2') }}
```

This will fail in 2030! jk jk

.github/workflows/manual-test.yaml
```diff
@@ -0,0 +1,76 @@
...
+  workflow_dispatch:
+    inputs:
+      conda-version:
+        description: 'Conda Version'
+        required: true
+        default: '23.1.0'
+        type: choice
+        options:
+        - 'main'
+        - '23.1.0'
```
jaimergp, line 13: I wouldn't give options here, but simply document you can use versions or a branch name.

.github/workflows/manual-test.yaml
```diff
@@ -0,0 +1,76 @@
...
+        shell: bash -el {0}
+        run: |
+          git clone https://github.com/conda/conda-libmamba-solver
+          cd conda-libmamba-solver
+          git checkout ${{ inputs.conda-libmamba-version }}
+          which conda
+          conda install -y -n base \
+             --file $PWD/dev/requirements.txt \
+             --file $PWD/tests/requirements.txt
+          python -m pip install -e . --no-deps
```
jaimergp, line 61: We need the initialization stuff too I think.

`python -m conda init --dev --all` maybe?

jaimergp:
@costrouc Are we still pursuing this? I am going to mark it as a draft for now.

costrouc:
Yeah mark as draft I may even just close this

PR #65, opened 2022-11-14
Load all channels with conda internals

jaimergp:
### Description

Closes https://github.com/conda/conda-libmamba-solver/issues/59

---

We are replacing the libmamba networking stack with conda internals:

- We download all JSONs with the newest `conda.SubdirData.repo_fetch` interface that doesn't process the JSON (only downloads). 
- We use `libmambapy.Repo`, without the `SubdirData` indirection. This interface allow us to use a JSON file directly (without having to be optimistic about the "find the JSON file" logic in `SubdirData`).
- All the tests that checked the libmamba networking stack have been removed, but the utilities are still there in case we ever need them in the future.
- These changes also allow us to drop tons of pre- and post-processing of MatchSpec and PackageRecord objects (e.g. percentage-encoding and decoding on local paths so libcurl doesn't choke, Windows things, etc). The libmamba<->conda index interface is better abstracted now.

---

There's some legacy issues we should remove in the future:

- The `SubdirData` subclass can introduce some issues with incomplete cache entries we have to invalidate (some tests break otherwise), because metaclass-based caches are that fun.
- In some tests, the cached files are not found because the integration tests abuse the `SubdirData._cache_` a bit too much and some of those changes are not reflected on disk. Some changes upstream will alleviate this.
- TODO: 
  - [ ] Change `conda.testing.helpers._patch_for_local_exports` to _not_ patch `_mtime` to `inf`. Return early if the solver is classic because that function is not needed there.
  - [ ] Remove [this block](https://github.com/conda/conda-libmamba-solver/blob/739dd05b0ab1abe523bf8535e3c3f34c556345e6/conda_libmamba_solver/index.py#L229-L234)

PR #149, opened 2023-03-08
try to use purpose-built download API

dholth:
### Description

See https://github.com/conda/conda/pull/12461 which was designed to improve cases like this. Should we continue to go through properties on SubdirData()?


jaimergp:
This is now https://github.com/conda/conda/pull/12475, right? I'll keep an eye on that PR to see if we can bring it over to #65

jaimergp:
Superseded by latest changes in #65. Thanks @dholth!

PR #103, opened 2022-12-24
Simplify exception parsing

jaimergp:
### Description

Investigating a simpler workflow for exception catching, in preparation the new error messages (see #102).

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #157, opened 2023-03-14
Investigate linux py310 cache not loaded error

jaimergp:
### Description

Looking into this error we have started seeing today in several CI jobs:

https://github.com/conda/conda-libmamba-solver/pull/103#issuecomment-1468409816

### Checklist - did you ...


jaimergp:
Of course it involves some kind of side effect 😂 It doesn't reproduce if you only run `tests/test_solvers.py`... Let's see which one it is.

jaimergp:
Reproduced in https://github.com/conda/conda-libmamba-solver/actions/runs/4420440414/jobs/7750132757#step:4:1133 and locally.

Turns out the env var was wrongly set up. So that's the fix. An extra S.

jaimergp:
@conda/conda-libmamba-solver: Ready for review :) 

Note this is also fixed by #65 (no libmamba cache involved), but for now we are just fixing the test itself.

jezdez (approved):

tests/test_solvers.py
```diff
@@ -113,8 +113,13 @@ def test_determinism(tmpdir):
     # - Input order (from the user side) should not matter, and should give the same result
     for i, pkg_list in enumerate(chain(repeat(pkgs, 10), permutations(pkgs, len(pkgs)))):
         offline = ("--offline",) if i else ()
-        out = check_output([*common_args, *offline, *pkg_list], env=env)
-        data = json.loads(out)
+        process = run([*common_args, *offline, *pkg_list], env=env, text=True, capture_output=True)
+        if process.returncode:
+            print("Attempt:", i)
+            print(process.stdout)
+            print(process.stderr, file=sys.stderr)
```
costrouc, line 120: Like this since it will make this test verbose when it does fail.

tests/test_solvers.py
```diff
@@ -93,7 +93,7 @@ def test_determinism(tmpdir):
     "Based on https://github.com/conda/conda-libmamba-solver/issues/75"
     env = os.environ.copy()
     env.pop("PYTHONHASHSEED", None)
-    env["CONDA_PKGS_DIR"] = str(tmpdir / "pkgs")
+    env["CONDA_PKGS_DIRS"] = str(tmpdir / "pkgs")
```
costrouc, line 96: Interesting. Wonder how this was passing in the past.

costrouc (approved):

Issue #71, opened 2022-02-21
`conda update` behaves differently with `libmamba` solver


jaimergp:

beenje:
I ran into the same issue:

```
$ conda create -y -n myapp python=3.10 flask
$ conda list -n myapp | grep flask
flask                     2.2.2              pyhd8ed1ab_0    conda-forge

$ conda update -n myapp --quiet  --solver classic flask
# All requested packages already installed.

$ conda update -n myapp --quiet  --solver libmamba flask
The following packages will be DOWNGRADED:

  flask                                  2.2.2-pyhd8ed1ab_0 --> 2.2.1-pyhd8ed1ab_0
```

This is with conda 22.11.1 and conda-libmamba-solver 22.12.0

costrouc:
@beenje that is a really helpful example. I was able to reproduce. To help here are the libsolve instructions which might help to resolve. I've been spending some time trying to better understand the logic that https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/solver.py#L332 does which is the state presented to libsolve.

```shell
conda create -y -n myapp python=3.10 flask
```

```
task: INSTALL 256
 - flask
 - python=3.10
```

```
$ conda list -n myapp | grep flask
flask                     2.2.3              pyhd8ed1ab_0    conda-forge
```

```
$  conda update -n myapp --quiet  --solver libmamba flask
```

Stage 1

```
USERINSTALLED 2560
 - conda-forge/noarch::flask==2.2.3=pyhd8ed1ab_0
 - conda-forge/linux-64::python==3.10.9=he550d4f_0_cpython
 - conda-forge/linux-64::ca-certificates==2022.12.7=ha878542_0
 - conda-forge/linux-64::openssl==3.1.0=h0b41bf4_0
task: UPDATE | ESSENTIAL 131840
 - flask!=2.2.3
 - conda-forge/linux-64::python==3.10.9=he550d4f_0_cpython
 - ca-certificates
 - openssl
```

```
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    flask-2.2.2                |     pyhd8ed1ab_0          74 KB  conda-forge
    ------------------------------------------------------------
                                           Total:          74 KB

The following packages will be DOWNGRADED:

  flask                                  2.2.3-pyhd8ed1ab_0 --> 2.2.2-pyhd8ed1ab_0 
```

PR #158, opened 2023-03-23
Minor change to test how this affects the rest of CI

costrouc:
Closes https://github.com/conda/conda-libmamba-solver/issues/156
Closes https://github.com/conda/conda-libmamba-solver/issues/71

### Description

This PR aims to fix issues when a user run `conda update ...` and in turn conda-libmamba-solver recommends to downgrade the users packages. This PR involves investigating the spec <-> tasks translations and how these are fed into libsolv.

### Checklist - did you ...


costrouc:
I was able to demonstrate how setting `>=...` did fix the issue shown in #71 but I'm curious how this affects other tests.

costrouc:
Current failing tests:
 - tests/test_create.py::IntegrationTests::test_create_only_deps_flag https://github.com/conda/conda-libmamba-solver/actions/runs/4496472144/jobs/7911184496?pr=158#step:5:1611 (several times)
 - tests/gateways/test_connection.py::test_s3_server https://github.com/conda/conda-libmamba-solver/actions/runs/4496472144/jobs/7911181839?pr=158#step:9:1075 (several times)

I can reproduce the first one but on the s3 one.

conda_libmamba_solver/solver.py
```diff
@@ -453,6 +453,10 @@ def _specs_to_tasks_add(self, in_state: SolverInputState, out_state: SolverOutpu
                             "UPDATE | ESSENTIAL | FORCEBEST",
                             api.SOLVER_UPDATE | api.SOLVER_ESSENTIAL | api.SOLVER_FORCEBEST,
                         )
+                    elif self._command == "update":
```
jaimergp, line 456: Hm, we already checked for this in line 445 above. I wonder if we can simply change the `!=` in the else below? I remember experimenting with this but something else broke. And we also need to check for the error messages. If a package cannot be updated because the latest version is already there, then that's not an error per se.

jaimergp:
> I can reproduce the first one but not the s3 one.

The S3 one is unrelated. I've seen it in other PRs. I think `minio` released a new server or something and we need to pin.

costrouc:
Link to tests on d6c66ca
 - https://github.com/conda/conda-libmamba-solver/actions/runs/4673486979
 - https://github.com/conda/conda-libmamba-solver/actions/runs/4673486981

jaimergp:
Now we have some more clarity on the failing tests:

```
FAILED ../conda-libmamba-solver-src/tests/test_modified_upstream.py::test_freeze_deps_1
FAILED ../conda-libmamba-solver-src/tests/test_modified_upstream.py::test_conda_downgrade
```

These are upstream tests that we vendored to modify the expectations a bit. There's a chance we can readjust them again, but also it might be that the `>=` changes introduce this regression too. To be investigated.

cc @costrouc

costrouc:
@jaimergp getting on this right now. Thanks for rebasing to rerun the CI.

jaimergp:
So this might be working now! The only thing that changed is that libmamba chooses to keep one dependency around (`xz`) while conda classic prunes it. In the past we could workaround that in the test by adding `prune=True` to the solver, but not anymore. So I removed that modification and commented out the `xz` line. It's sufficiently edge-casey (and keeping a package is not too critical of an error) so we can ignore it, imho.

conda_libmamba_solver/solver.py
```diff
@@ -262,7 +262,10 @@ def _solving_loop(
                 "as conflicts for maximum flexibility",
             )
             # we only check this for "desperate" strategies in _specs_to_tasks
-            self._command = "last_solve_attempt"
+            if self._command in (None, NULL):
```
jezdez, line 265: zilch, nada!

jezdez (approved):

PR #172, opened 2023-03-30
Workaround for `defaults::*` specs and `solver.explain_problems()` crashes

jaimergp:
### Description

```
$ CONDA_SOLVER=libmamba conda create -n dssfdjaime -c defaults python pip defaults::libarchive
Channels:
 - defaults
 - conda-forge
Platform: linux-aarch64
Collecting package metadata (repodata.json): done
Solving environment: / error    libmamba Selected channel specific (or force-reinstall) job, but package is not available from channel. Solve job will fail.
- error    libmamba Selected channel specific (or force-reinstall) job, but package is not available from channel. Solve job will fail.
\ warning  libmamba Problem type not implemented SOLVER_RULE_JOB_UNSUPPORTED
python: /home/conda/feedstock_root/build_artifacts/mamba-split_1679568211070/work/libmamba/src/core/satisfiability_error.cpp:1628: mamba::{anonymous}::TreeExplainer::write_leaf(const mamba::{anonymous}::TreeNode&)::<lambda(const auto:43&)> [with auto:43 = mamba::ProblemsGraph::RootNode]: Assertion `false' failed.
Aborted
```

This crashes Python, no traceback. We are requesting something wrong to libsolv (working on a fix), but then `explained_problems()` crashes trying to elaborate. We anticipate for those errors and avoid the call.

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -385,7 +385,7 @@ class LibMambaSolver(Solver):
         for name, spec in out_state.specs.items():
             if name.startswith("__"):
                 continue
-            self._check_spec_compat(spec)
+            spec = self._check_spec_compat(spec)
             spec_str = self._spec_to_str(spec)
             installed = in_state.installed.get(name)
 
@@ -491,7 +491,7 @@ class LibMambaSolver(Solver):
         # --deps-only
         key = ("ERASE | CLEANDEPS", api.SOLVER_ERASE | api.SOLVER_CLEANDEPS)
         for name, spec in in_state.requested.items():
-            self._check_spec_compat(spec)
+            spec = self._check_spec_compat(spec)
             tasks[key].append(str(spec))
 
         return dict(tasks)
@@ -504,7 +504,7 @@ class LibMambaSolver(Solver):
         for name, spec in out_state.specs.items():
             if name.startswith("__"):
                 continue
-            self._check_spec_compat(spec)
+            spec = self._check_spec_compat(spec)
             spec = self._fix_version_field_for_conda_build(spec)
             tasks[key].append(spec.conda_build_form())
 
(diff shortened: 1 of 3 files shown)


conda_libmamba_solver/solver.py
```diff
@@ -618,7 +621,18 @@ def _maybe_raise_for_problems(
...
-        return f"{self.solver.problems_to_str()}\n{self.solver.explain_problems()}"
+        legacy_errors = self.solver.problems_to_str()
+        if "unsupported request" in legacy_errors:
+            # This error makes 'explain_problems()' crash. Anticipate.
+            log.info("Failed to explain problems. Unsupported request")
+            return legacy_errors
+        try:
+            explained_errors = self.solver.explain_problems()
+        except Exception as exc:
+            log.info("Failed to explain problems", exc_info=exc)
```
jezdez, line 632: Would it be better here to use `log.error` or `log.warning`?

tests/test_solvers.py
```diff
@@ -89,6 +89,37 @@ def test_python_downgrade_reinstalls_noarch_packages():
...
+def test_defaults_specs_work():
+    """
+    Reported in https://github.com/conda/conda/issues/11346
+
+    See also test_create::test_noarch_python_package_reinstall_on_pyver_change
+    in conda/conda test suite. Note that we use conda-forge here deliberately;
+    defaults at the time of writing (March 2022) packages pip as a non-noarch
+    build, which means it has a different name across Python versions. conda-forge
+    uses noarch here, so the package is the same across Python versions. Probably
+    why upstream didn't catch this error before.
```
jezdez, line 101: We're not using pip here anymore to test this, but libarchive, should we update the docstring?

conda_libmamba_solver/solver.py
```diff
@@ -741,6 +755,24 @@ def _check_spec_compat(self, match_spec):
...
+            # We need to replace it with the actual channel name (main, msys2, r)
+            # Instead of searching in the index, we apply a simple heuristic:
+            # - R packages are [_]r-*, mro-*, rpy or rstudio
+            # - Msys2 packages are m2-*, m2w64-*, or msys2-*
+            # - Everything else is in main
+            name = match_spec.name.lower()
+            if name in ("r", "rpy2", "rstudio") or name.startswith(("r-", "_r-", "mro-")):
+                channel = "pkgs/r"
+            elif name.startswith(("m2-", "m2w64-", "msys2-")):
+                channel = "pkgs/msys2"
```
jezdez, line 770: Yikes, this hard-coding is worrying, can we get that list from somewhere?

jezdez (commented):
Just

conda_libmamba_solver/solver.py
jaimergp: Not without querying the `index` and summarising ourselves. This is just a temporary patch to enable `defaults::*` functionality, which is broken on `libmamba` right now. I am hoping that this is fixed by the next release so we can remove this. I don't think `msys2` and `r` see too many changes these days anyway, and the number of people doing `defaults::r` should be very small (they would just do `-c defaults r` if anything).

tests/test_solvers.py
jaimergp: Ah no, this is a copy-paste error. removing, thanks!

conda_libmamba_solver/solver.py
jaimergp: This will pollute the stdout with no real value to the user. If they miss the explanation, they can enable verbosity to see why. We shouldn't ever reach this part though so, yea, let's do it. I'll go for warning.

conda_libmamba_solver/solver.py
jezdez: For the record, the upstream PR is https://github.com/conda/conda/pull/12555

jezdez (approved):

PR #191, opened 2023-04-24
Adding libmamba/libmambapy as development dependencies

costrouc:
Closes #190 

This work adds documentation on how to install mamba and include it as a development dependency.

That said this extra work adds in total 10 min with ~5 min solving for libmamba/libmambapy dependencies and ~5 min of compilation.

PR #193, opened 2023-04-24
Upcoming changes in mamba context breaks conda-libmamba-solver

### LINKED ISSUE #192: Upcoming changes in mamba context breaks conda-libmamba-solver
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The following commit to mamba https://github.com/mamba-org/mamba/pull/2432/ changes the way in which the context object is constructed. This means the next libmambapy version will break conda-libmamba-solver. Will be opening a PR on the fixes but the bigger question is how to preserve compatibility with older libmamba versions.

We are going to have to continue with repodata patches potentially due to this? Hope this issue/PR opens a larger discussion.

```python
ctx = libmambapy.context.Context()
ctx.json # Attribute Error
ctx.output_config.json # exists 
```

costrouc:
Closes #192 

The following commit to mamba https://github.com/mamba-org/mamba/pull/2432/ changes the way in which the context object is constructed. This means the next libmambapy version will break conda-libmamba-solver. Will be opening a PR on the fixes but the bigger question is how to preserve compatibility with older libmamba versions.

We are going to have to continue with repodata patches potentially due to this? Hope this issue/PR opens a larger discussion.

```python
ctx = libmambapy.context.Context()
ctx.json # Attribute Error
ctx.output_config.json # exists
```

This PR should not be merged since mamba with these changes has not been released yet.

This was possible since I was able to test upstream mamba commits via PR #191 

This is expected to fail on versions `<=1.4.2` since there has not been a release yet with #2432.


costrouc:
@jaimergp since conda-libmamba-solver doesn't depend on mamba for networking logic. Should we even be setting the network parameters in context?

jaimergp:
Yea we can probably get away without touching the networking stuff. We have all those entries because they were matching the original `utils.py` in `mamba-org/mamba` (in their pre-1.0 days), but these days we have diverged enough to consider writing this function from scratch.

In the past we have dealt with `context` API changes by testing with `hasattr`. So for the new stuff I'd write something like:

```python
if hasattr(ctx, "output_config"):
    # set all the output stuff
else:
    # set it in the global namespace
```

For the old conda-libmamba-solver versions, we'll have to repodata patch to limit the upper bound, no way around that 🤷

jaimergp:
Note libmamba 1.4.3 was cut _without_ the deprecation wrappers, but 1.4.4 immediately followed adding them. None have been released so far in conda-forge or defaults so we are still safe.

jezdez:
What's the latest on this, do we need to do anything about that?

jaimergp:
We only need a review and a test with 1.4.4, now that is out.

jezdez (approved):

PR #248, opened 2023-07-27
libmamba returns URL-encoded paths that break our channel info retrieval

jaimergp:
### Description

Closes https://github.com/conda/conda-libmamba-solver/issues/247

Local tests reveal that this issue is not present in libmamba 1.4.1.

### Checklist - did you ...

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import json
 import os
+import shutil
+import sys
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


jaimergp:
As seen locally, libmamba 1.4.1 (from defaults) is not affected by the bug. libmamba 1.4.9 (only available in conda-forge) does hit the problem. You can see this in the `Tests / *` CI. I am sending the fix now that we have public record of the issue.

jaimergp:
Failures are unrelated (same as we can see in #245)

kenodegard (approved):

PR #242, opened 2023-07-20
Adding a new channel results in too many package updates

jaimergp:
### Description

Will close #240 

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #266, opened 2023-08-24
Investigate compatibility with libmamba 1.5.0

jaimergp:
### Description

One test broke with libmamba 1.5.0: `test_solvers.py::test_defaults_specs_work`. libmamba is not able to handle `defaults::*` as a matchspec, but in <=1.4.9 we could workaround by translating defaults to the most likely `pkgs/*` equivalent. This workaround doesn't work in 1.5.0 anymore so I am simply xfailing the test and adding a cleaner exception to the UI.

This is reported upstream at https://github.com/mamba-org/mamba/issues/2431, but no solution is available yet.

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -364,7 +364,21 @@ class LibMambaSolver(Solver):
         tasks = self._specs_to_tasks(in_state, out_state)
         for (task_name, task_type), specs in tasks.items():
             log.debug("Adding task %s with specs %s", task_name, specs)
-            self.solver.add_jobs(specs, task_type)
+            try:
+                self.solver.add_jobs(specs, task_type)
+            except RuntimeError as exc:
+                if mamba_version() == "1.5.0":
+                    for spec in specs:
+                        if spec in str(exc):
+                            break
+                    else:
+                        spec = f"One of {specs}"
+                    msg = (
+                        "This is a bug in libmamba 1.5.0. 'defaults' (and 'pkgs/*') cannot be "
+                        "used in channel::pkg_name specs. Please use -c defaults instead."
+                    )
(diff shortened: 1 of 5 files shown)


jaimergp:
Errors are unrelated (see #267 and #242). This doesn't fix the underlying issue but does annotate the offending test as xfail. Work is needed upstream for a proper fix.

conda_libmamba_solver/solver.py
```diff
@@ -364,7 +364,21 @@ def _solve_attempt(
...
+                if mamba_version() == "1.5.0":
+                    for spec in specs:
+                        if spec in str(exc):
+                            break
+                    else:
+                        spec = f"One of {specs}"
+                    msg = (
+                        "This is a bug in libmamba 1.5.0. 'defaults' (and 'pkgs/*') cannot be "
+                        "used in channel::pkg_name specs. Please use -c defaults instead."
+                    )
```
jezdez, line 379: While technically correct, is this really something we want to bubble up to users?

```suggestion
                    msg = (
                        "This is a bug in libmamba 1.5.0 when using 'defaults::"
                        "<spec>' or 'pkgs/main::<spec>'. Please use '-c "
                        " defaults' instead."
                    )
```

Is `pkgs/main::<spec>` correct?

jezdez (approved):
Just minor nit

conda_libmamba_solver/solver.py
jaimergp:
```suggestion
                    msg = (
                        "This is a bug in libmamba 1.5.0 when using 'defaults::"
                        "<spec>' or 'pkgs/main::<spec>'. Please use '-c "
                        "defaults' instead."
                    )
```

PR #270, opened 2023-08-28
Alternative logic for solver creation tasks

jaimergp:
### Description

Same as #242 but trying other methods.

### Checklist - did you ...


jezdez:
@jaimergp I concur, this sounds like a good idea. We might be able to package libmamba 1.5.0 ahead of the release as well, of course, to retrofit testing on defaults as well.

jaimergp:
Some good progress thanks to the pin fixes in `main` via https://github.com/mamba-org/mamba/pull/2786. Note the flags we were using (see e40f0da3c96651312fa6dc765aaf259836ff12df) are not compatible with `solver.add_pin`, so I had to remove them. Doesn't seem to make a difference so far in `conda/conda`'s `test_solve.py`.

Let's see what the CI says! 👀

jaimergp:
Ok, not bad, we are getting closer:

Upstream:
- [ ] `tests/cli/test_main_clean.py::test_clean_tempfiles`
- [ ] `tests/cli/test_subcommands.py::test_create`
- [ ] `tests/cli/test_subcommands.py::test_env_create`
- [ ] `tests/core/test_solve.py::test_only_deps_2`
- [ ] `tests/test_create.py::test_install_features`
- [ ] `tests/test_export.py::test_explicit`
- [ ] `tests/test_create.py::test_update_all_updates_pip_pkg`
- [ ] `tests/test_export.py::test_export`
- [ ] `tests/test_create.py::test_package_pinning`

Unit:
- [ ] `tests/test_modified_upstream.py::PatchedCondaTestCreate::test_neutering_of_historic_specs`
- [ ] `tests/test_modified_upstream.py::test_update_all_1`
- [ ] `tests/test_modified_upstream.py::test_python2_update`

costrouc:
@jaimergp if I understand this PR correctly the idea is to move more in line with what mamba tasks are created and simplify the logic. As a side effect it makes this easier to reason about and debug?

jaimergp:
Hi @costrouc, welcome back! Yes, your assessment is correct. I think we have an all-green CI now if you look at conda-forge items running libmamba 1.5.0. There are some failures but they do pass with a hypothetical `1.5.1` running from `main `. That said, we are still skipping some tests that I think we might be able to tackle now with the cleaner logic.

So, what I'd like to do here is:

- Please review this PR thoroughly and ask as many questions as you need.
- Take a couple hours to investigate what's wrong with the tests that involve `accelerate`, `iopro`, `mkl` and so on. We consistently get slightly different behaviours and I suspect is because of the (deprecated) `features` functionality, which libmamba doesn't fully process. Some examples of these tests (currently deselected, so uninstall the pytest plugin):

- `tests/core/test_solve.py::test_features_solve_1`
- `tests/core/test_solve.py::test_prune_1`
- `tests/core/test_solve.py::test_update_prune_2`
- `tests/core/test_solve.py::test_update_prune_3`


What I've seen too is that this logic is more correct than the previous one, in the sense that it passes some tests from `test_modified_upstream` _only_ if we revert to the original. So, in essence, I'd like to revisit our skipped / xfailed modules and see which ones are now passing. Can you take a look? Thanks! Remember that this needs `libmamba` from current `main`.

news/267-prune
```diff
@@ -0,0 +1,21 @@
+### Enhancements
+
+* <news item>
```
jaimergp, line 3: Needs updating.

jaimergp:
@costrouc I'd rather not force push to this branch. I cannot see which changes you have submitted, and the starting commit hash is not mentioned anywhere in this history. Can you clarify what you were trying to change? Thanks.

jaimergp:
libmamba 1.5.1 should be out on defaults too

costrouc:
@jaimergp yes this was my bad. I accidentally pushed to this branch and force pushed to remove the commit that I added.

You can clobber anything that I have done in the branch. There should not have been any changes.

Can you revert back to https://github.com/conda/conda-libmamba-solver/commit/670dd2f9eac75cf5a87114cfb32fdf74a7c39e6e

costrouc:
I've also tested `tests/test_create.py::test_neutering_of_historic_specs` and it still fails with the latest on mamba.

pyproject.toml
```diff
@@ -23,8 +23,8 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-  "conda >=23.5.0",
-  "libmambapy >=1.4.1",
+  "conda >=23.7.3",
+  "libmambapy >=1.5.0",
```
jaimergp, line 27:
```suggestion
  "libmambapy >=1.5.1",
```

recipe/meta.yaml
```diff
@@ -21,8 +21,8 @@ requirements:
     - hatch-vcs
   run:
     - python >=3.8
-    - conda >=23.5.0
-    - libmambapy >=1.4.1
+    - conda >=23.7.3
+    - libmambapy >=1.5.0
```
jaimergp, line 25:
```suggestion
    - libmambapy >=1.5.1
```

jaimergp:
`tests/test_misc.py::test_explicit_missing_cache_entries` fails because `PackageCacheData.get_all_extracted_entries()[0].url` is None (on macOS only!). I can't reproduce locally, it's not solver-related, and it passes on other operating systems 🤷

jezdez (approved):

PR #289, opened 2023-09-21
Do not allow pin overrides via requested specs in the CLI

jaimergp:
### Description

Closes #286 

The original intention of this PR was to address an issue in how we were checking whether pinned specs and explicitly requested specs were compatible or not. After careful investigation, we realized two things:

- ~generalizing a robust solution would require querying the index in several occasions trying to anticipate something that the solver will anyway end up doing. This is RAM intensive and potentially more costly than the solve it self~. It's not that bad.
- This check is part of the "user-requested specs can override pinned specs" logic implemented in the classic solver. This behaviour has been addressed as surprising in issue https://github.com/conda/conda/issues/9016.

As a result, in this PR we have decided to:

* Prevent the user from overriding pinned specs with incompatible CLI specs. An early error will show up if that happens. Instead, we suggest being more explicit and adjust the pinned specs, or use `--no-pin`.
* Formalize a "name-only pins lock if installed" policy. This adds support for pinned specs that are name-only (no version or build constraints). This was previously undefined behavior. For every name-only pinned spec:
  * If the package is installed, we lock it. It won't be able to be change at all.
  * If it's not installed, we don't do anything.
  * If the user needs to modify that package, again, the pinned config needs to be modified explicitly or `--no-pin` can be used.
  * The error message will render name-only pins as their installed counterparts if appropriate.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #294, opened 2023-09-26
Allow only CLI specs as long as compatible with pins

jaimergp:
### Description

Extends #289 to allow CLI specs compatible with pins. It will still reject incompatible ones, with the standard `SpecsConfigurationConflictError`.

### Checklist - did you ...


jaimergp:
Some profiling reveals that querying the index is not that much of an overhead (see `index.py:search`):

image

The big picture shows it's not a major problem:

image

jezdez (approved):

PR #295, opened 2023-09-26
use .solv cache if available and recent

jaimergp:
### Description

`libmambapy.Repo` takes a local JSON file to load their internal metadata. We were passing the JSON path, but it's faster to pass their SOLV cache (from ~4s to ~2s in my conda-forge/linux-aarch64 repodata; might be faster for bigger ones like linux-64). We only use it if it's more recent that the JSON copy.

When there's no SOLV cache:

image

If we use the SOLV cache:

image

We save a couple seconds there.

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/index.py
+++ b/conda_libmamba_solver/index.py
@@ -74,6 +74,7 @@ import logging
 import os
 from dataclasses import dataclass
 from functools import lru_cache, partial
+from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Dict, Iterable, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


jaimergp:
Adding this now because I noticed it while I was analyzing the profiles at #294.

costrouc:
Tested out this PR and got the following results. No warnings or error raised.

`time conda create -n test numpy -c conda-forge --dry-run`
 - First run (no repodata) 37.4s
 - Second run 32.0s
 - Third run 32.1s

(no solv files) in cache (checked)

`time conda create --solver libmamba -n test numpy -c conda-forge --dry-run`
 -Fourth run 41.8s
 - Fifth run 5.0s

`rm ~/.conda/pkgs/cache/*.solv`
`time conda create --solver libmamba -n test numpy -c conda-forge --dry-run`

 - Sixth run 14.3s

So this using .solv file takes off about 9s for me. I was able to repeat these results.

costrouc (approved):

PR #298, opened 2023-09-27
Speed up `_notify_conda_outdated` due to expensive PrefixData call.

costrouc:
### Description

Profiling conda-libmamba-solver after the changes that @jaimergp made in https://github.com/conda/conda-libmamba-solver/pull/295 I got the following results. Without using cProfile (since profiling slows things down) 5.05s on my hardware (with nvme drive and repodata cached)

`time conda create --solver libmamba --dry-run -n test numpy -c conda-forge`

Here is the profile data collected with `snakeviz`

image

Notice that `_notify_conda_outdated` is taking 2s/5.41s total and it approximately 1/3 of the time it takes for the command to run. When looking at the profiling data `PrefixData(..).get('conda', None)` is expensive at `0.08ms` for each call to `_load_single_record(...)` in total 222 calls were made totalling those 2s. All just to fetch the `conda` PrefixRecord.

I explored two approaches:
 - Make PrefixData(...).get('conda', None)` lazy so it doesn't actually read every record
 - Manually read the conda prefixrecord

I worked toward implementing both and things looked fine when making prefix data lazy BUT then implementing `PrefixData._load_site_packages` as lazy was not simple and didn't look like it would actually speed things up once this function is implemented. This is due to `_load_site_packages` calling `self.iter_records()` https://github.com/conda/conda/blob/main/conda/core/prefix_data.py#L278 which in effect makes all the work to make prefixdata lazy worthless. So we are better off implementing our own.

So with this implementation we get `4.29s` for the runtime about a 15% speedup (with slower drives the speedup will be more). With the following profile

image

Out of curiousity I compared this with `mamba` (which uses PrefixData so we would expect it to be slower). I ensuring that caching is used `5.98s`. Then I compared with micromamba getting around `2.6s`. So I would estimate that conda's overhead compared with a compiled implementation is around 1-2s as of now using conda-libmamba-solver.

### Checklist - did you ...


costrouc:
Made a fix to check that the conda-meta directory actually exists forgot that part of the `PrefixData.load(..)` implementation.

conda_libmamba_solver/solver.py
```diff
@@ -1004,8 +1004,15 @@ def _notify_conda_outdated(
             # This check can be silenced with a specific option in the context or in quiet mode
             return
 
-        current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
-        if not current_conda_prefix_rec:
+        # manually check base prefix since `PrefixData(...).get("conda", None) is expensive
+        # once prefix data is lazy this might be a different situation
+        current_conda_prefix_rec = None
+        for filename in os.scandir(os.path.join(context.conda_prefix, "conda-meta")):
+            if filename.name.endswith(".json") and filename.name.rsplit("-", 2)[0] == "conda":
```
jaimergp: I think `pathlib.Path.glob()` uses scandir under the hood, so maybe we can get away with the more readable:


```suggestion
        for filename in Path(context.conda_prefix, "conda-meta").glob("conda-*.json"):
            if filename.name.rsplit("-", 2)[0] == "conda":
```

jaimergp: Ah, only on recent Python versions. Ugh!

conda_libmamba_solver/solver.py
```diff
@@ -1004,8 +1004,17 @@ def _notify_conda_outdated(
             # This check can be silenced with a specific option in the context or in quiet mode
             return
 
-        current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
-        if not current_conda_prefix_rec:
+        # manually check base prefix since `PrefixData(...).get("conda", None) is expensive
+        # once prefix data is lazy this might be a different situation
+        current_conda_prefix_rec = None
+        conda_meta_prefix_directory = os.path.join(context.conda_prefix, "conda-meta")
+        if os.path.lexist(conda_meta_prefix_directory):
```
jaimergp, line 1011:
```suggestion
        if os.path.lexists(conda_meta_prefix_directory):
```

Typo? And why `lexists` instead of `exists`?

costrouc: Yeup fixed!

jaimergp:
Errors are either expected (#296) or connection errors. This is good to go.

conda_libmamba_solver/solver.py
```diff
@@ -1004,7 +1005,20 @@ def _notify_conda_outdated(
...
+        with suppress(OSError):
+            if os.path.lexists(conda_meta_prefix_directory):
+                for filename in os.scandir(conda_meta_prefix_directory):
+                    if (
+                        filename.name.endswith(".json")
+                        and filename.name.rsplit("-", 2)[0] == "conda"
+                    ):
+                        with open(filename.path) as f:
+                            current_conda_prefix_rec = PrefixRecord(**json.load(f))
+                            break
```
jezdez, line 1021:
```suggestion
                for entry in os.scandir(conda_meta_prefix_directory):
                    if (
                        entry.is_file()
                        and entry.name.endswith(".json")
                        and entry.name.rsplit("-", 2)[0] == "conda"
                    ):
                        with open(entry.path) as entry_file:
                            current_conda_prefix_rec = PrefixRecord(**json.load(entry_file))
                            break
```

It's better to clarify that `os.scandir` returns a list of entries, not filenames, and you have to check whether the entry is a file or not.

I wonder what the `json.load` call might trigger in terms of exceptions, which would bubble up. It would be good to be a little more defensive about this.

jezdez (changes requested):

conda_libmamba_solver/solver.py
dholth: It might not matter for tiny-JSON but FYI json.loads(entry_file.read()) is faster than passing a file object to json.load()

jezdez: Ah good point, lemme update my suggestion

jezdez: done

conda_libmamba_solver/solver.py
```diff
@@ -1004,7 +1005,20 @@ def _notify_conda_outdated(
             # This check can be silenced with a specific option in the context or in quiet mode
             return
 
-        current_conda_prefix_rec = PrefixData(context.conda_prefix).get("conda", None)
+        # manually check base prefix since `PrefixData(...).get("conda", None) is expensive
+        # once prefix data is lazy this might be a different situation
+        current_conda_prefix_rec = None
+        conda_meta_prefix_directory = os.path.join(context.conda_prefix, "conda-meta")
+        with suppress(OSError):
```
jezdez, line 1012:
```suggestion
        with suppress(OSError, ValueError):
```

`json.JSONDecodeError` is a subclass of `ValueError`, so we'll catch json decoding issues as well

jezdez (approved):

conda_libmamba_solver/solver.py
```diff
@@ -1004,7 +1005,20 @@ def _notify_conda_outdated(
...
+        with suppress(OSError, ValueError):
+            if os.path.lexists(conda_meta_prefix_directory):
+                for entry in os.scandir(conda_meta_prefix_directory):
+                    if (
+                        entry.is_file()
+                        and entry.name.endswith(".json")
+                        and entry.name.rsplit("-", 2)[0] == "conda"
+                    ):
+                        current_conda_prefix_rec = PrefixRecord(**json.loads(entry_file.read()))
+                            break
```
jezdez, line 1021:
```suggestion
                        break
```

jezdez (approved):

conda_libmamba_solver/solver.py
costrouc: Thanks for this feedback everyone. Didn't know that scandir returned this much. In the future I think I will use this method much more heavily.

PR #323, opened 2023-10-13
Fail gracefully if offline and no cache is available

jaimergp:
### Description

Needed for https://github.com/conda/conda/pull/12984

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/index.py
+++ b/conda_libmamba_solver/index.py
@@ -76,7 +76,7 @@ from dataclasses import dataclass
 from functools import lru_cache, partial
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Dict, Iterable, Tuple, Union
+from typing import Dict, Iterable, Tuple, Union, Optional
 
 import libmambapy as api
 from conda.base.constants import REPODATA_FN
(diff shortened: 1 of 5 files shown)


costrouc (approved):
Tested and can confirm in `pytest tests/test_create.py -k test_offline_with_empty_index_cache\[libmamba\]` that the test continues but still does eventually fail. The error messages though are more helpful in debugging.

costrouc:
I will be investigating the tests that fail and why it is happening reliably on windows.

jezdez (approved):

costrouc:
Triggering a rerun and seeing what happens. I ran this test locally on a windows vm and wasn't able to reproduce (which is expected since this is a conda activation error).

costrouc:
Spent a long time trying to debug the windows issue. I see a difference of `assert PATH0.lower() == PATH.lower()` in left side having `/usr/bin` in the path and the other has `/bin` in the path. On line https://github.com/conda/conda/blob/e784a57001b129804f17e10cce4cd31400be8103/tests/test_activate.py#L2696.

jaimergp:
`tests/conda_env/test_cli.py::test_update_env_no_action_json_output` errors are legit but caused by a unique situation. This is the gist:

* Having an `env.yml` file with `python=3`, `pip` and `pip: click`, we should be able to create it with `conda env create`.
* A subsequent call to `conda env update` should do nothing because we just installed everything.

Makes sense, but this is what we observe in this error:
1. `conda env create` creates an environment with `python=3.11` and `pip=23.3`. It prioritizes having a greater pip version than a Python one, possibly because it comes earlier in the sorting (i beats y).
2. However, `conda env update` imposes `python=3` again (not just `python`!), which is interpreted as "update `python`". This time, the solver tries to satisfy the petition more aggressively, even allowing a `pip` downgrade this time, because the alternative is to say "we are ok".

We didn't see this error earlier because the packaging situation is unique: `defaults` has Python 3.12 now, but only `pip` 23.2 is available for that version. The most recent 23.3 is only available for Python 3.11. I'm assuming once the new pip is available for 3.12 this will pass, but the currently observed behavior is a bit surprising and might deserve a fix.

jaimergp:
Note that changing the `environment.yml` file to the following makes everything pass as expected:

```diff
ENVIRONMENT_PYTHON_PIP_CLICK = f"""
name: {TEST_ENV_NAME_1}
dependencies:
- - python=3
+ - python
  - pip
  - pip:
    - click
channels:
  - defaults
"""
```

jaimergp:
This alternative change in `solver.py` also makes things pass. `MatchSpec.merge` sorts everything by `str`, and we used to keep that:

```python
        self.specs_to_add = IndexedSet(MatchSpec.merge(s for s in fixed_specs))
```

However, if we prio Python, it passes:

```python
        resorted_specs = []
        for spec in MatchSpec.merge(s for s in fixed_specs):
            if spec.name == "python":
                resorted_specs.insert(0, spec)
            else:
                resorted_specs.append(spec)
        self.specs_to_add = IndexedSet(resorted_specs)
```

PR #339, opened 2023-10-27
Update recipe to attempt to fix local build failure

dholth:
### Description

I wasn't able to conda-build the embedded recipe, not because of noarch. This works better. Still having trouble with hatch, hatch-vcs, and pip's idea of what a valid version number is however.

Also, call pthread_sigmask . I tested this on OSX and you can't press CTRL-C at all without it. It's still unreliable to interrupt download threads but that is being worked on in https://github.com/conda/conda/pull/13246

### Checklist - did you ...

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,0 +1,57 @@
+package:
+  name: "conda-libmamba-solver"
+  version: "{{ GIT_DESCRIBE_TAG }}.{{ GIT_BUILD_STR }}"
+
+source:
+  # git_url is nice in that it won't capture devenv stuff.  However, it only
+  # captures committed code, so pay attention.
+  git_url: ../
+  folder: src/
+
+build:
+  number: 0
+  skip: true  # [py<38]
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ GIT_DESCRIBE_TAG }}.{{ GIT_BUILD_STR }}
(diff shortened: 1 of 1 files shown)


jaimergp:
Hi Daniel! Why is this needed? We are mimicking conda-forge noarch recipe on purpose. 

If this is for development, I think we have some instructions in the docs.

dholth:
@jaimergp this is needed because "conda-build ." fails on a checkout of the main branch.

beeankha:
pre-commit.ci autofix

jezdez:
It's worth separating the packaging related proposal here into another PR.

jaimergp:
I don't think there's anything wrong with the recipe. `conda build recipe/` should work correctly in a clean workspace (e.g. `CONDA_BLD_PATH=~/tmp/new-build-workspace conda build recipe/`).

_However_, I've noticed that we will error with traceback due to a bad argument in the exception chain. This can be reproduced if the conda-build local channel contains packages in the dependency tree that would shadow the remote channels (e.g. if you have built a `conda` package recently, that's in your `CONDA_BLD_PATH` channel, so it will shadow your remote `defaults` or `conda-forge` `conda` package due to strict priority). When that happens, we error out with a solver conflict, _but_ the exception is built with a missing argument, and it fails badly. I'll open a separate PR to address that, but I'm closing this one.

PR #347, opened 2023-10-31
Require libmambapy 1.5.3

jaimergp:
### Description

This [release](https://github.com/mamba-org/mamba/releases/tag/libmambapy-1.5.3) includes a few fixes that we have needed in conda-libmamba-solver for a while, including:

- Compatibility with `defaults::pkg` and `pkgs/main::pkg` specs (https://github.com/mamba-org/mamba/pull/2936)
- Ability to bypass the C++ level signal handling (https://github.com/mamba-org/mamba/pull/2946)

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/mamba_utils.py
+++ b/conda_libmamba_solver/mamba_utils.py
@@ -88,25 +88,20 @@ def init_api_context() -> api.Context:
     api_ctx = api.Context()
 
     # Output params
-    # We use this getattr() trick to guarantee backwards compatibility
-    # with libmambapy <=1.4.2; 1.4.3+ grouped some settings together
-    output_params = getattr(api_ctx, "output_params", api_ctx)
-    output_params.json = context.json
-    output_params.quiet = context.quiet
-    output_params.verbosity = context.verbosity
+    api_ctx.output_params.json = context.json
+    api_ctx.output_params.quiet = context.quiet
+    api_ctx.output_params.verbosity = context.verbosity
     api_ctx.set_verbosity(context.verbosity)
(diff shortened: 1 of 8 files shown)


jaimergp:
Note that libmamba 1.5.3 is not on defaults yet, but the CI seems to be getting it from 1.5.3 from conda-forge anyway (even on the defaults CI entries), hence why everything's passing.

conda_libmamba_solver/mamba_utils.py
```diff
@@ -88,25 +88,20 @@ def init_api_context() -> api.Context:
...
-    output_params = getattr(api_ctx, "output_params", api_ctx)
-    output_params.json = context.json
-    output_params.quiet = context.quiet
-    output_params.verbosity = context.verbosity
+    api_ctx.output_params.json = context.json
+    api_ctx.output_params.quiet = context.quiet
+    api_ctx.output_params.verbosity = context.verbosity
     api_ctx.set_verbosity(context.verbosity)
-    if output_params.json:
+    if api_ctx.output_params.json:
```
dholth, line 95: We are cancelling some other json if json is True?

kenodegard (approved):

conda_libmamba_solver/mamba_utils.py
jaimergp: We don't want libmamba to print JSON on its own. It's something only conda should do.

jezdez (approved):

jaimergp:
Merging with failures because they are connection errors.

PR #378, opened 2023-11-13
Always sort installed records and solver specs

jaimergp:
### Description

I've been thinking about some failures we get (only!) sometimes in CI, like `tests/core/test_solve.py::test_conda_downgrade[libmamba]` (see #317). These errors often go away with a retry in the same PR, so there's some randomness at play. 

The hypothesis here is that some container is getting a different order depending on the attempt. `SolverInputState.installed` takes whatever order appears in `PrefixData._prefix_records`, which blindly "takes" whatever `os.scandir()` is returning. This function does not guarantee order, so we are going to switch to a toposorted `installed` container. Maybe this makes the tests behave more reliably?

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -407,8 +407,8 @@ class LibMambaSolver(Solver):
             tasks = self._specs_to_tasks_conda_build(in_state, out_state)
         else:
             tasks = self._specs_to_tasks_add(in_state, out_state)
-        print(
-            "Created following tasks:\n",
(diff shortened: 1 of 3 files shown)


jaimergp:
Seems to work. The errors we saw in the last run (first attempt) are unrelated (#334, and a pip issue with `classic` that smells like `rm_rf` race condition).

See https://github.com/conda/conda-libmamba-solver/issues/317 for the whole debugging story.

jezdez (approved):

PR #394, opened 2023-11-21
More explicit logic for CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS

jaimergp:
### Description

Inspired by #391. Refactoring a bit before we delve in the question of "how many attempts should we try before giving up?".

cc @mbargull

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -36,6 +36,7 @@ from conda.common.url import join_url, percent_decode
 from conda.core.prefix_data import PrefixData
 from conda.core.solve import Solver
 from conda.exceptions import (
+    CondaValueError,
     InvalidMatchSpec,
     InvalidSpec,
     PackagesNotFoundError,
@@ -256,6 +257,35 @@ class LibMambaSolver(Solver):
                 return "Getting pinned dependencies"
         return "Solving environment"
 
+    def _max_attempts(self, in_state: SolverInputState):
+        from_env_var = os.environ.get("CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS")
+        n_installed = len(in_state.installed)
+        if from_env_var:
+            try:
+                max_attempts_from_env = int(from_env_var)
+            except ValueError:
+                raise CondaValueError(
+                    f"CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='{from_env_var}'. Must be int."
+                )
+            if max_attempts_from_env < 1:
+                raise CondaValueError(
+                    f"CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='{max_attempts_from_env}'. Must be >=1."
+                )
+            if max_attempts_from_env > n_installed:
+                log.warning(
+                    "CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='%s' is higher than the number of "
+                    "installed packages (%s) and will be ignored.",
+                    max_attempts_from_env,
+                    n_installed,
+                )
+                return n_installed
+            return max_attempts_from_env
+        if in_state.update_modifier.FREEZE_INSTALLED:
+            # this the default, but can be overriden with --update-specs
+            # TODO: should we cap this at a reasonable number? some base envs have 100s of pkgs
+            return n_installed
+        return 1
+
     def _solving_loop(
         self,
         in_state: SolverInputState,
@@ -263,11 +293,7 @@ class LibMambaSolver(Solver):
         index: LibMambaIndexHelper,
     ):
         solved = False
-        max_attempts = max(
-            2,
-            int(os.environ.get("CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS", len(in_state.installed))) + 1,
-        )
-        for attempt in range(1, max_attempts):
+        for attempt in range(1, self._max_attempts(in_state) + 1):
             log.debug("Starting solver attempt %s", attempt)
             try:
                 solved = self._solve_attempt(in_state, out_state, index, attempt=attempt)
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -282,8 +282,8 @@ class LibMambaSolver(Solver):
             return max_attempts_from_env
         if in_state.update_modifier.FREEZE_INSTALLED:
             # this the default, but can be overriden with --update-specs
-            # TODO: should we cap this at a reasonable number? some base envs have 100s of pkgs
-            return n_installed
+            # we cap at 10 attempts to avoid things getting too slow in large environments
+            return min(10, n_installed)
         return 1
 
     def _solving_loop(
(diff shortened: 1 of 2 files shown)


conda_libmamba_solver/solver.py
```diff
@@ -256,18 +257,43 @@ def _spinner_msg_solving(self):
...
+                    "CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='%s' is higher than the number of "
+                    "installed packages (%s) and will be ignored.",
+                    max_attempts_from_env,
+                    n_installed,
+                )
+                return n_installed
+            return max_attempts_from_env
+        if in_state.update_modifier.FREEZE_INSTALLED:
+            # this the default, but can be overriden with --update-specs
+            # TODO: should we cap this at a reasonable number? some base envs have 100s of pkgs
```
jaimergp, line 285: This is the main point of this PR :) Thoughts?

mbargull: I haven't messed with `libmamba` solver at all, so absolutely no idea/feeling about what ranges would be sensible.

If you don't want a hard limit, you could use some function that just increases slower for higher inputs.
If you want a hard limit you could either do `min(n_installed, limit)` or something that only approaches the limit gradually like `math.ceil(n_installed / (1 + (n_installed / limit)**2)**.5)` or something more/less sophisticated.

jaimergp: Each solve is usually sub-second, but some complex one that need some extra backtracking (like the one reported in your issue) take a 2-4s. If you end up with 100 installed packages then you need to wait a few minutes for the solver to give up.

By "giving up" I mean that we stop trying to unlock installed packages and we let the solver modify them as needed, just as mamba does. So maybe we can get by with 10 attempts. In other words, let the solver unfreeze up to 5-10 conflicting installed packages one by one until we just let everything float.

Alternatively, we could make the retry loop be based on time spent and not iterations, but that might be more complex than necessary.

jaimergp: I'm going to try with 10 attempts and see if that is enough to pass all tests. Then we adjust as necessary.

mbargull: Makes sense.
Could you add that as a constant variable (`DEFAULT_LIBMAMBA_SOLVER_MAX_ATTEMPTS` or something ) to better track that?

----
> Alternatively, we could make the retry loop be based on time spent and not iterations, but that might be more complex than necessary.

Yes, complexity, but also determinism ;).

jaimergp:
Errors are due to https://github.com/conda/conda/issues/13360

conda_libmamba_solver/solver.py
```diff
@@ -256,18 +258,44 @@ def _spinner_msg_solving(self):
...
+                )
+                return n_installed
+            return max_attempts_from_env
+        if in_state.update_modifier.FREEZE_INSTALLED:
+            # this the default, but can be overriden with --update-specs
+            # we cap at MAX_SOLVER_ATTEMPTS_CAP attempts to avoid things
+            # getting too slow in large environments
+            return min(self.MAX_SOLVER_ATTEMPTS_CAP, n_installed)
+        return 1
+
```
jezdez, line 290:
```suggestion
    def _max_attempts(self, in_state: SolverInputState, default: int = 1):
        from_env_var = os.environ.get("CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS")
        installed_count = len(in_state.installed)
        if from_env_var:
            try:
                max_attempts_from_env = int(from_env_var)
            except ValueError:
                raise CondaValueError(
                    f"CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='{from_env_var}'. Must be int."
                )
            if max_attempts_from_env < 1:
                raise CondaValueError(
                    f"CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='{max_attempts_from_env}'. Must be >=1."
                )
            elif max_attempts_from_env > installed_count:
                log.warning(
                    "CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS='%s' is higher than the number of "
                    "installed packages (%s). Using that one instead.",
                    max_attempts_from_env,
                    installed_count,
                )
                return installed_count
            else:
                return max_attempts_from_env
        elif in_state.update_modifier.FREEZE_INSTALLED:
            # this the default, but can be overriden with --update-specs
            # we cap at MAX_SOLVER_ATTEMPTS_CAP attempts to avoid things
            # getting too slow in large environments
            return min(self.MAX_SOLVER_ATTEMPTS_CAP, installed_count)
        else:
            return default

```

conda_libmamba_solver/solver.py
```diff
@@ -256,18 +258,44 @@ def _spinner_msg_solving(self):
...
         out_state: SolverOutputState,
         index: LibMambaIndexHelper,
     ):
         solved = False
-        max_attempts = max(
-            2,
-            int(os.environ.get("CONDA_LIBMAMBA_SOLVER_MAX_ATTEMPTS", len(in_state.installed))) + 1,
-        )
-        for attempt in range(1, max_attempts):
+        for attempt in range(1, self._max_attempts(in_state) + 1):
```
jezdez, line 298: Would it be useful to prepopulate the max attempts at the start of the solve as a instance attribute, instead of inline, so it's easier to debug?

jezdez (commented):
Just a few suggestions, nothing major

mbargull:
> @mbargull To clarify, does this actually fix #391 for you or is this just a nice-to-have?

No, gh-391 actually failed on the first attempt for me.
The example from gh-391 was just one of those for which @jaimergp noticed long run times for when it didn't segfaulted (since in that example it has to downgrade some packages).

----
gh-381 incidentally fixes the example from gh-391, but it is unclear if other cases might still fail depending on the ordering; see the discussing from https://github.com/conda/conda-libmamba-solver/issues/391#issuecomment-1820660838 onward.

conda_libmamba_solver/solver.py
jaimergp: I'm not sure if this might break some tests. It shouldn't, but by saving it as instance attribute we are assuming we are going to have the same number of attempts for every call to `solve_for_*()` (for the lifetime of the same instance). In some instances, that might be tied to `len(in_state.installed)`.

So I'd rather keep it like it is because it's technically more correct and does not assume that "one instantiation, one solve".

jezdez:
> > @mbargull To clarify, does this actually fix #391 for you or is this just a nice-to-have?
> 
> No, [gh-391](https://github.com/conda/conda-libmamba-solver/issues/391) actually failed on the first attempt for me. The example from [gh-391](https://github.com/conda/conda-libmamba-solver/issues/391) was just one of those for which @jaimergp noticed long run times for when it didn't segfaulted (since in that example it has to downgrade some packages).
> 
> [gh-381](https://github.com/conda/conda-libmamba-solver/pull/381) incidentally fixes the example from [gh-391](https://github.com/conda/conda-libmamba-solver/issues/391), but it is unclear if other cases might still fail depending on the ordering; see the discussing from [#391 (comment)](https://github.com/conda/conda-libmamba-solver/issues/391#issuecomment-1820660838) onward.

Ah, thanks for the pointer, much appreciated!

conda_libmamba_solver/solver.py
jezdez: Sounds good, thank you!

PR #421, opened 2024-01-09
Refine `MatchSpec` validation to avoid incompatibilities with libmamba's parser

jaimergp:
### Description

Closes https://github.com/conda/conda-libmamba-solver/issues/418

We were a bit too strict with some auto-generated `MatchSpec`. The issue above includes an environment where a user has installed a package via URL or path. In this case, `conda` generates a `MatchSpec` object that gets rendered as `package-name[url="package-url"]`.

libmamba only supports _some_ key values in the brackets, so we were rejecting `url` with a `InvalidSpecError`. However, `url` here is superfluous so we can simply drop it from the passed string.

This uncovered a new matchspec that makes libmamba segfault: `<unknown>/noarch::test-package`. We will also anticipate that and drop it before the string is generated because, again, it doesn't add any useful info for the solver and it's only an artifact of not knowing from which channel test-package was installed (because it came from a URL).

Note that _if_ the URL happens to be a channel URL (e.g. `conda.anaconda.org/conda-forge/noarch/package-version-build.tar.bz2`), `conda` DOES recognize it as a valid channel and that doesn't crash. It only reproduces _if_ the tarball is in a non-channel-like location.

### Checklist - did you ...

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -13,7 +13,7 @@ from conda.common.compat import on_linux, on_win
 from conda.common.io import env_vars
 from conda.core.prefix_data import PrefixData
 from conda.models.channel import Channel
-from conda.testing.integration import _get_temp_prefix, make_temp_env
+from conda.testing.integration import _get_temp_prefix, make_temp_env, package_is_installed
 from conda.testing.integration import run_command as conda_inprocess
 
 from .channel_testing.helpers import http_server_auth_basic  # noqa: F401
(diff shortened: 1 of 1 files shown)


jaimergp:
Failing as expected. Submitting fix now.

therc:
Would this also fix the error

`InvalidMatchSpec: Invalid spec 'conda-forge/linux-64::_libgcc_mutex==0.1=conda_forge[md5=d7c89558ba9fa0495403155b64376d81]': Libmamba only supports a subset of the MatchSpec interface for now. You can only use ('name', 'version', 'build', 'channel', 'subdir'), but you tried to use ('md5',).`

assuming that "md5" would/could be dropped, too?

jaimergp:
Hm, maybe not (yet). We are only dropping the url field. Can you open a new issue so we can take a look? It's weird that the md5 hash ends up in the matchspec.

nhsavage:
new issue opened at #427

yzhang-gh:
May I ask whether this fix is included in the latest version? I have updated conda to v24.9.2 as suggested in https://conda.github.io/conda-libmamba-solver/user-guide/#how-to-install but still got this error

```
InvalidMatchSpec: Invalid spec 'pytorch3d[url=file:///home/xxx/pytorch3d-0.7.8-py310_cu121_pyt211.tar.bz2]':
Libmamba only supports a subset of the MatchSpec interface for now.
You can only use ('name', 'version', 'build', 'channel', 'subdir'), but you tried to use ('url',).
```

jaimergp:
What's your `conda-libmamba-solver` version? Maybe we don't support `url` in the field, but you can use it directly.

yzhang-gh:
Thanks for the reply. I tried `conda list -n base conda-libmamba-solver` and it shows 23.12.0.

Could you elaborate a little bit about how to use it directly? I installed the `pytorch3d` package `conda install --offline` and after that I can no longer use `conda install` to install other packages.

jaimergp:
Try to upgrade to conda-libmamba-solver 24.9 first, which should contain this fix. If that doesn't work, report back.

yzhang-gh:
That's it! Many thanks.

PR #457, opened 2024-03-26
Refactor for libmamba v2 (WIP)

jaimergp:
### Description

Adds compatibility for libmamba v2, which enables a few refactors (see below). Supersedes conda/conda#414.

- Will close https://github.com/conda/conda-libmamba-solver/issues/522

Progress:

- [X] Refactor index helpers
- [X] Refactor repoquery
- [x] Refactor solver
- [x] Refactor conflict reports
- [x] Pass all tests
- [ ] Polish and clean up

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

### Summary of changes

- Deprecate `CONDA_LIBMAMBA_SOLVER_NO_CHANNELS_FROM_INSTALLED`. No longer needed with v2. Closes conda/conda#411
- Added typing for all methods and functions
- Added `time_recorder` decorator for metadata collection and solving loop
- Removed `state.BaseIndexHelper`. We just have our `index.IndexHelper` now
- The libmamba v1 `pool` (collection of `repos`, where a `repo` represents a loaded `repodata.json`) is now a `database` of `RepoInfo` objects.
- Do use `current_repodata.json` if explicitly set in CLI
- Move to Ruff for pre-commit linting & formatting
- Logging from libsolv has a HUGE overhead now because it goes from C to C++ to Python to logging to stdout instead of C -> stdout. To enable libsolv verbosity, an extra environment variable `CONDA_LIBMAMBA_SOLVER_DEBUG_LIBSOLV` is required.
- `allow_uninstall` was previously set to `false` (only `true` for `conda remove`), and we set it up for individual solver jobs involving updates or conflicts. With v2, we have individual control over what to "Keep" instead of drop. This required marking important updates as keepers instead. Otherwise they would be uninstalled.
- Other changes in the test suite discussed in https://github.com/conda/conda/pull/13784

PR #482, opened 2024-07-01
Do try to load .solv files by default

### LINKED ISSUE #481: .solv files not being reused
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

It looks like in the following commit https://github.com/conda/conda-libmamba-solver/commit/8d9d6c03fd13e6a86a16fc396e1f7413baa269ea

a `try_solv` argument was added to `_json_path_to_repo_info` which defaults to `False`.  

From what I can tell this was added so that `reload_local_channels` could invalidate the `.solv` files but only other call to `_json_path_to_repo_info` in `_load_channels` did not get updated to use `try_solv=True`.  

This is causing every solve attempt to parse the repodata json and regenerate the `.solv` files even when all the channels are cache hits.  This leads to a performance hit.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

jaimergp:
### Description

Closes #481 

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/index.py
+++ b/conda_libmamba_solver/index.py
@@ -256,7 +256,7 @@ class LibMambaIndexHelper(IndexHelper):
         return url, json_path
 
     def _json_path_to_repo_info(
-        self, url: str, json_path: str, try_solv: bool = False
+        self, url: str, json_path: str, try_solv: bool = True
     ) -> _ChannelRepoInfo | None:
         channel = Channel.from_url(url)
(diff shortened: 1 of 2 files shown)


jaimergp:
Errors here are also present in main and the fix is included in https://github.com/conda/conda-libmamba-solver/pull/457, so I think there's no need to spend cycles here given how simple the change is.

jezdez (approved):

PR #586, opened 2024-12-03
Type annotation in LibMambaSolver

dholth:
### Description

Noticed error when passing Path.

Alternatively, change it to str instead of os.PathLike to match superclass.

### Checklist - did you ...

### DIFF
--- a/news/accept-pathlike
+++ b/news/accept-pathlike
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Accept `Path()` for `LibMambaSolver()` prefix, not just str.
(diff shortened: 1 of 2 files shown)


news/accept-pathlike
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Accept `Path()` for `LibMambaSolver()` prefix, not just str.
```
jaimergp, line 7:
```suggestion
* Accept both `pathlib.Path()` and `str` for `LibMambaSolver()` prefix. (#586)
```

jaimergp (approved):

PR #610, opened 2025-01-28
test: Use libmambapy 2.0.6.rc3

jjerphan:
### Checklist - did you ...

### DIFF
--- a/dev/requirements.txt
+++ b/dev/requirements.txt
@@ -3,5 +3,5 @@ pip
 # run-time
 boltons>=23.0.0
 conda>=23.7.3
-conda-forge/label/mamba_prerelease::libmamba==2.0.6.rc2
-conda-forge/label/mamba_prerelease::libmambapy==2.0.6.rc2
+conda-forge/label/mamba_prerelease::libmamba==2.0.6.rc3
+conda-forge/label/mamba_prerelease::libmambapy==2.0.6.rc3
--- a/dev/requirements.txt
+++ b/dev/requirements.txt
@@ -3,5 +3,5 @@ pip
 # run-time
 boltons>=23.0.0
 conda>=23.7.3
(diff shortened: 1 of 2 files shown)


jjerphan:
There are errors because the root prefix cannot be determined.

`MAMBA_ROOT_PREFIX` can be set to do so. Previously it was always set thanks to a profile script wrongly shipped in `libmamba`, but this changed with https://github.com/mamba-org/mamba/pull/3723.

Now: should we set it instead of the user or not?

jaimergp:
Sure, we can patch it at runtime and set it before importing `libmamba` (assuming the env var set in `os.environ` is seen by C++?). I am more intrigued by the need for that value. If it can't be detected, shouldn't the library detect its own location and pick the relevant parent directory? Or even better: should it be specified explicitly as a parameter in the necessary API calls instead of depending on an environment variable?

jjerphan:
Currently the location of `mamba` (which might not be installed) is used to locate the root prefix if the environment variable is not set, but we could use the location of `libmamba` instead.

dev/requirements.txt
```diff
@@ -3,5 +3,5 @@ pip
 # run-time
 boltons>=23.0.0
 conda>=23.7.3
-conda-forge::libmamba>=2.0.0
-conda-forge::libmambapy>=2.0.0
+conda-forge/label/mamba_prerelease::libmamba==2.0.6.rc1
+conda-forge/label/mamba_prerelease::libmambapy==2.0.6.rc1
```
jjerphan, line 7:
```suggestion
conda-forge/label/mamba_prerelease::libmamba==2.0.6.rc2
conda-forge/label/mamba_prerelease::libmambapy==2.0.6.rc2
```

jjerphan:
Closing as 2.0.6.rc3 is successful (the only failure is due to a timeout).

Issue #613, opened 2025-02-04
KeyError in post_solve when using --only-deps


beenje:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Running `conda create -n sardana --only-deps sardana-core sardana-config taurus-qt` raises the following exception:

```
(base) root@940120b469ff:/# conda create -n sardana --only-deps sardana-core sardana-config taurus-qt
Channels:
 - conda-forge
Platform: linux-aarch64
Collecting package metadata (repodata.json): done
Solving environment: failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/opt/conda/lib/python3.12/site-packages/conda/exception_handler.py", line 18, in __call__
        return func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.12/site-packages/conda/cli/main.py", line 61, in main_subshell
        exit_code = do_call(args, parser)
                    ^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.12/site-packages/conda/cli/conda_argparse.py", line 205, in do_call
        result = getattr(module, func_name)(args, parser)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.12/site-packages/conda/notices/core.py", line 132, in wrapper
        return func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.12/site-packages/conda/cli/main_create.py", line 154, in execute
        return install(args, parser, "create")
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.12/site-packages/conda/cli/install.py", line 409, in install
        unlink_link_transaction = solver.solve_for_transaction(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.12/site-packages/conda/core/solve.py", line 153, in solve_for_transaction
        unlink_precs, link_precs = self.solve_for_diff(
                                   ^^^^^^^^^^^^^^^^^^^^
[...]

PR #618, opened 2025-02-18
Do not raise if record is already removed (e.g. with --deps-only)

jaimergp:
### Description

Closes https://github.com/conda/conda-libmamba-solver/issues/613

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/state.py
+++ b/conda_libmamba_solver/state.py
@@ -692,7 +692,8 @@ class SolverOutputState:
                 if installed:
                     self.records[name] = installed
                 else:
-                    self.records.pop(record.name)
+                    # Record might not be included with flags like --only-deps; ignore in that case
+                    self.records.pop(record.name, None)
 
         elif sis.update_modifier.UPDATE_DEPS:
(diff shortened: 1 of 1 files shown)


jaimergp:
pre-commit.ci autofix

jaimergp:
I had to add a new block of code that essentially contradicts this comment:

```python
# Using a special instance of PrefixGraph to remove youngest child nodes that match
# the original requested specs.  It's important to remove only the *youngest* child
# nodes, because a typical use might be `conda install --only-deps python=2 flask`,
# and in that case we'd want to keep python.
#
# What are we supposed to do if flask was already in the environment?
# We can't be removing stuff here that's already in the environment.
#
# What should be recorded for the user-requested specs in this case? Probably all
# direct dependencies of flask.
```

I think we should actually NOT keep python, but I special cased it. My intuition as a user would be like `conda install --only-deps A B C` shouldn't leave any of A, B, C in the final solution, even if C depends on B. The user made the effort of writing it down, so it must be important. If they really need to condition the solver with more constraints, then they should do it in more steps or, you know, `conda remove --force` their way there.

PR #686, opened 2025-08-15
Support sharded repodata by creating a subset of reachable packages in Python, and then sending that subset to conda-libmamba-solver.

### LINKED ISSUE #684: Modify LibMambaIndexHelper for Shard-Based Repodata Gathering
### Checklist

### What is the idea?

Change the way LibMambaIndexHelper gathers repodata in conda-libmamba-solver's Python code. Instead of downloading complete repodata files, it should look for shards and compute repodata subsets by traversing all existing and requested dependencies.

### Why is this needed?

The current approach of downloading entire repodata files is inefficient and slow, especially when only a subset of packages is actually needed for dependency resolution. A shard-based approach would significantly reduce download times and bandwidth usage.

### What should happen?

The LibMambaIndexHelper class needs to be refactored to detect and work with sharded repodata instead of monolithic files. It should implement logic to identify which shards are needed based on the dependency graph and only fetch those specific shards.

### Additional Context

_No response_

### LINKED ISSUE #706: Add Repodata Shards Detection with ETag Support
**What is the idea?**
Implement functionality to check for the existence of repodata_shards.msgpack.zst files and handle caching based on ETag values to avoid unnecessary downloads.

**Why is this needed?**
Efficient caching requires the ability to determine whether cached data is still valid. ETag support allows the system to quickly check if remote shards have been updated without downloading the entire file.

**What should happen?**
Add code to detect when repositories support sharded repodata by looking for repodata_shards.msgpack.zst files. Implement ETag comparison logic to return cached data when ETags match the last fetch, or trigger new downloads when they differ.

**Additional Context**
The implementation should handle repositories that don't support shards gracefully and fall back to traditional repodata handling. Network error handling and timeout scenarios should be considered.

dholth:
### Description

Working on adding sharded repodata support, mostly in the Python code of conda-libmamba-solver

Fix #684 
Fix #706

### Checklist - did you ...


conda_libmamba_solver/index.py
```diff
@@ -160,6 +162,7 @@ def __init__(
         repodata_fn: str = REPODATA_FN,
         installed_records: Iterable[PackageRecord] = (),
         pkgs_dirs: PathsType = (),
+        in_state: SolverInputState | None = None,
```
jaimergp, line 236: Looks like this is not being used, is it? I'd rather pass the bits you need to the methods that require it so these three objects (Index, InputState, OutputState) do not depend on each other

dholth: You may have a better suggestion, but, the idea is that we would include the shard fetching before line 186 `# XXX needs to be lazy (to not download "classic" repodata eagerly)` if `in_state is not None`.
In an earlier commit I made `self.repos` lazy but it caused weird errors in the test suite or when running conda in the debugger.

tests/channel_testing/reposerver.py
```diff
@@ -343,11 +343,14 @@ def do_GET(self):
...
     handler = CondaTokenHandler
+else:
+    print("No auth method given.")
+    exit(1)
 
 PORT = args.port
 
 server = HTTPServer(("", PORT), handler)
-print("Server started at localhost:" + str(PORT))
+print("Server started at localhost:" + str(server.server_port))
```
dholth, line 353: Updating this to allow an OS-chosen port (ask for port 0, and the OS chooses an available port).

travishathaway: Nice trick!

conda_libmamba_solver/shards_subset.py
```diff
@@ -0,0 +1,126 @@
...
+    subset_paths = {}
+
+    repodata_size = 0
+    for channel, shardlike in channel_data.items():
+        repodata = shardlike.build_repodata()
+        # XXX not guaranteed unique
+        _, *channel_shortname = channel.rsplit("/", 2)
+        repodata_path = tmp_path / ("_".join(channel_shortname))
+        # most compact json
+        repodata_text = json.dumps(repodata, indent=0, separators=(",", ":"))
```
jaimergp:
```suggestion
        repodata_text = json.dumps(repodata, indent=0, separators=(",", ":"), sort_keys=True)
```

conda_libmamba_solver/shards_subset.py
```diff
@@ -0,0 +1,126 @@
...
+
+def build_repodata_subset(tmp_path, root_packages, channels):
+    channel_data: dict[str, ShardLike] = {}
+    for channel in channels:
+        for channel_url in Channel(channel).urls(True, context.subdirs):
+            subdir_data = SubdirData(Channel(channel_url))
+            found = fetch_shards(subdir_data)
+            if not found:
+                repodata_json, _ = subdir_data.repo_fetch.fetch_latest_parsed()
+                found = ShardLike(repodata_json, channel_url)  # type: ignore
```
jaimergp: I was reviewing https://github.com/conda/conda-libmamba-solver/pull/697, and I wondered why do we need a ShardLike interface? Then saw this code. Are we going to subset all repodata going forward? That will remove any benefits from the .solv cache in Unix systems that just memory map the file and introduce maybe unneeded overhead? Also in general, reading the full JSON just to subset it, and then write it back, only to be read again by libmamba... I guess it's faster to just pass it through.

So... what if we do _not_ use `ShardLike` for non-sharded channels and just keep the original repodata.json?

dholth: I would rather parse them into mamba PackageRecord and skip json, that's possible right?
I have read blog posts saying memory mapping isn't that big a win on a lot of modern systems.
Using the original cached json path instead of a fresh subset-of-monolithic-repodata would be a simple choice when loading LibMambaIndexHelper; even though to subset the real sharded repodata we also need to compute the subset of fake sharded repodata.
You'd be surprised at how fast Python json.loads() runs, what we must avoid is creating PackageRecord for all the unused dependencies.

conda_libmamba_solver/shards_subset.py
dholth: making sure everything's in order

conda_libmamba_solver/shards_subset.py
jaimergp: We would still need to initialize a bunch of Python objects to pass them to libmamba. My point is that if (in the case of non sharded repodata) we already have a JSON on disk ready to pass to the libmamba constructors, why bother doing all these extra steps.

jaimergp: For sharded repodata, we could avoid the JSON writing overhead and construct the Python objects for libmamba straight from the raw dict. And then measure if that's faster. I guess there's a sweet number somewhere we can use as a threshold if the benefits are not clear in all cases.

dholth: I hope you noticed the part where we don't do any of this if we have no sharded channels.

conda_libmamba_solver/index.py
```diff
@@ -282,7 +294,22 @@ def _load_channels(
             encoded_urls_to_channel[url] = channel
         urls_to_channel = encoded_urls_to_channel
 
-        urls_to_json_path_and_state = self._fetch_repodata_jsons(tuple(urls_to_channel.keys()))
+        if self.in_state:
```
dholth: vs. out of state

tests/data/in_state.pickle
dholth, line 1: Includes a pickled posix Path object that can't be deserialized on Windows

conda_libmamba_solver/shards.py
dholth: profiler shows this line takes a significant amount of time

conda_libmamba_solver/shards.py
```diff
@@ -325,58 +337,89 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
...
     fetch = sd.repo_fetch
-    cache = fetch.repo_cache
+    repo_cache = fetch.repo_cache
     # cache.load_state() will clear the file on JSONDecodeError but cache.load()
     # will raise the exception
-    cache.load_state(binary=True)
-    cache_state = cache.state
+    repo_cache.load_state(
+        binary=True
+    )  # won't succeed when .msgpack.zst is missing as it wants to compare the timestamp (returns empty state)
```
dholth: This method clears the entire state if the timestamp on the cached file (either a msgpack.zst or a json) does not match a timestamp saved in the state file.
This is not helpful when all we want to do is check if the remote server had a certain file format last time we checked.
Instead, we should read the state file without checking it against any other file. This will tell us whether or not we should check the remote server for .msgpack.zst, or skip re-checking it for an interval.
If our locally cached file does not match the timestamp in the state file, we may only need to clear out the etag/other cache headers from the state. That ensures that we will not waste time checking for a remote file format that we know isn't there, and that we will create or replace the locally cached file if it is not consistent with the state file.

conda_libmamba_solver/shards.py
dholth: (resolved with _2 version of function)

conda_libmamba_solver/shards.py
dholth: For now we can load the json ourselves to get the should_check_format flag.

conda_libmamba_solver/shards.py
```diff
@@ -325,58 +341,106 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
...
+
+    wanted = []
+    for shard in sharded:
+        for package_name in packages:
+            if package_name in shard:
+                wanted.append((shard, package_name, shard.shard_url(package_name)))
+
+    log.debug("%d shards to fetch", len(wanted))
+
+    shared_shard_cache = sharded[0].shards_cache
```
jaimergp, line 452: This will raise `IndexError` if no shards were found in `sharded`.

conda_libmamba_solver/shards.py
```diff
@@ -98,6 +76,23 @@ def shard_mentioned_packages(shard: Shard) -> set[str]:
     return mentioned
 
 
+def shard_mentioned_packages_2(shard: ShardDict) -> Iterable[str]:
+    """
+    Return all dependency names mentioned in a shard, not including the shard's
+    own package name.
+    """
+    unique_specs = set()
+    for package in (*shard["packages"].values(), *shard["packages.conda"].values()):
```
jaimergp, line 67: Are these keys guaranteed to be present? Might want to `.get(..., {})` instead.

conda_libmamba_solver/shards.py
```diff
@@ -278,7 +291,10 @@ def fetch(s, url, package):
                             *result[fetch_result.package]["packages.conda"].values(),
```
jaimergp, line 301: Same here. Is the key guaranteed to exist?

conda_libmamba_solver/shards.py
dholth: Maybe we can write a short sanitizer near where we retrieve the shards

conda_libmamba_solver/shards.py
dholth: Worked around, but we might still subset normal repodata. We'll see if we can afford to do that, or not, once the tempfiles are eliminated.

conda_libmamba_solver/shards.py
dholth: I'm noticing that we do catch the KeyError below. Maybe this is a good place to insert validation, along with ShardLike which converts monolithic repodata into shards.

conda_libmamba_solver/shards.py
```suggestion
        # Should we accept a plain dict and validate here?
        self.repodata_no_packages: RepodataDict = {
            **repodata,
            "packages": {},
            "packages.conda": {},
        }
        all_packages: ShardDict = {}  # type: ignore
        for package_group in ("packages", "packages.conda"):
            package_group_packages = repodata.get(package_group, {})
            if not isinstance(package_group_packages, dict):
                raise TypeError(f"repodata {package_group} is not a dict")
            all_packages[package_group] = package_group_packages
            
```
Wasn't doing enough for me to want to commit

conda_libmamba_solver/shards.py
dholth: I notice we do iterate over all packages and check that they have names near where we insert shards into shardcache, and we do something similar to checking in ShardLike where we turn normal repodata into shards. We could promote these to raise ValueError if there is a problem.

conda_libmamba_solver/index.py
```diff
@@ -142,6 +151,72 @@ def canonical_name(self) -> str:
...
+    """
+    Flag to see whether we should check for sharded repodata.
+    """
+    return context.plugins.use_sharded_repodata is True  # type: ignore
+
+
+_SUPPORTS_PYTHON_SITE_PACKAGES = hasattr(PackageInfo, "python_site_packages_path")
+
+
+def _package_info_from_package_dict(
```
dholth, line 162: we might move the similar package info from PackageRecord out of the class, it doesn't use self either.

conda_libmamba_solver/index.py
```diff
@@ -282,7 +364,31 @@ def _load_channels(
...
+            # try to make a subset of possible dependencies
+            # conda probably already has a suitable temporary directory?
+            # keep a reference so GC doesn't delete the directory
+            self.tmp_dir = tempfile.TemporaryDirectory("conda-shards")
+            self.tmp_path = Path(self.tmp_dir.name)
+            root_packages = (*self.in_state.installed.keys(), *self.in_state.requested)
+            channel_data = build_repodata_subset(root_packages, urls_to_channel)
+            begin = time.monotonic_ns()
+            # XXX use this instead of urls_to_json_path_and_state; append directly to channel_repo_infos?
+            _subset_infos = self._load_repo_info_from_repodata_dict(channel_data)
```
dholth: If we are totally replacing `for url_w_cred, (json_path, state) in urls_to_json_path_and_state.items():` in shards mode, we could create the entire channel_repo_infos list here.
We probably preserve the channel order because of Python's ordered dict, to not worry about channel order.
We might want to add a sorted(package name keys) in _load_repo_info_from_repodata_dict "just in case"

conda_libmamba_solver/index.py
```diff
@@ -481,18 +587,54 @@ def _load_pkgs_cache(self, pkgs_dirs: PathsType) -> list[RepoInfo]:
...
+        for channel, shardlike in repodata_subset.items():
+            repodata = shardlike.build_repodata()
+            # Don't like going back and forth between channel objects and URLs;
+            # build_repodata_subset() expands channels into per-subdir URLs as
+            # part of fetch:
+            channel_object = Channel(channel)
+            packages = []
+
+            for package_group in ("packages", "packages.conda"):
+                for filename, record in repodata.get(package_group, {}).items():
```
dholth, line 600: option of removing .tar.bz2 duplicates of .conda here

conda_libmamba_solver/shards.py
dholth: [conda-index always produces these keys](https://github.com/conda/conda-index/blob/main/conda_index/index/sqlitecache.py#L374) for "real" shards. +1 on _fast_ sanitize, validate though.

Checking again in Shards, ShardLike, these keys are guaranteed by the time a shard gets to this point.

conda_libmamba_solver/shards.py
```diff
@@ -325,58 +375,180 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
...
+                    # not strictly true, since we could have fetched the same data
+                    # from repodata.json.zst; but makes the urljoin consistent with
+                    # shards which end with /repodata_shards.msgpack.zst
+                    url = f"{channel_url}/repodata.json"
+                    found = ShardLike(repodata_json, url)
+                channel_data[channel_url] = found
+        return channel_data
+
+    else:
+        # The parallel version may reorder channels, does this matter?
```
dholth, line 508: Could we make this as short as the serialized version?

conda_libmamba_solver/shards_subset.py
```diff
@@ -0,0 +1,155 @@
...
+        unvisited = [(n.distance, n) for n in self.nodes.values()]
+        sharded = [s for s in self.shardlikes if isinstance(s, Shards)]
+        to_retrieve: set[str] = set(self.nodes)
+        retrieved: set[str] = set()
+        while unvisited:
+            # parallel fetch all unvisited shards but don't mark as visited
+            if to_retrieve:
+                # can we get stuck looking for unavailable packages repeatedly?
+                not_in_cache = batch_retrieve_from_cache(sharded, sorted(to_retrieve))
+                batch_retrieve_from_network(not_in_cache)
```
dholth, line 130: Added network batching takes a "python, sqlalchemy, vaex" solve from 74s to 36s so far. The second time once all shards are cached, 5s.

conda_libmamba_solver/shards_subset.py
```diff
@@ -0,0 +1,155 @@
...
+                for package in shard_mentioned_packages_2(shard):
+                    if package not in self.nodes:
+                        self.nodes[package] = Node(node.distance + 1, package)
+                        # by moving yield up here we try to only visit dependencies
+                        # that no other node already knows about. Doesn't make it faster.
+                        if package not in discovered:  # redundant with not in self.nodes?
+                            log.debug(f"{json.dumps(node.package)} -> {json.dumps(package)};")
+                            yield self.nodes[package]
+                        else:
+                            assert False, "package already discovered but not in self.nodes"
```
dholth: This line is not executed by the unit tests. If we decide it is unreachable, we may remove it or mark it with _nocover_

conda_libmamba_solver/shards.py
```diff
@@ -124,10 +123,20 @@ def __init__(self, repodata: RepodataDict, url: str = ""):
...
 
         # used to write out repodata subset
-        self.visited: dict[str, Shard | None] = {}
+        self.visited: dict[str, ShardDict | None] = {}
+
+        # alternate location for packages, if not self.url
+        try:
+            base_url = self.repodata_no_packages["info"]["base_url"]
+            if not isinstance(base_url, str):
+                log.warning(f'repodata["info"]["base_url"] was not a str, got {type(base_url)}')
```
dholth, line 115: Not covered. We could add a base_url = 1 e.g. test case in the unit suite.

conda_libmamba_solver/shards.py
```diff
@@ -278,7 +322,10 @@ def fetch(s, url, package):
                             *result[fetch_result.package]["packages.conda"].values(),
                         )
                     ]
-                    log.debug("expected %s, actual %s", fetch_result.package, set(package_names))
+                    if set((fetch_result.package,)) != set(package_names):
```
dholth, line 304: Not covered. We could create a shard with multiple package names or the wrong package name for the unit suite.

conda_libmamba_solver/index.py
ryanskeith: It is nice to trim where we can.

conda_libmamba_solver/shards.py
```diff
@@ -256,17 +296,21 @@ def fetch(s, url, package):
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
             futures = {
                 executor.submit(fetch, self.session, url, package): (url, package)
                 for url, package in urls_packages.items()
                 if package not in result
             }
-            # May be inconvenient to cancel a large number of futures. Also, ctrl-C doesn't work reliably out of the box.
             for future in concurrent.futures.as_completed(futures):
                 log.debug(". %s", futures[future])
+                # XXX future.result can raise HTTPError etc.
```
ryanskeith, line 291: Do unhandled errors cause whole app crashes currently?

conda_libmamba_solver/shards.py
```diff
@@ -79,7 +59,7 @@ def ensure_hex_hash(record: PackageRecordDict):
     return record
 
 
-def shard_mentioned_packages(shard: Shard) -> set[str]:
+def shard_mentioned_packages(shard: ShardDict) -> set[str]:
```
ryanskeith: Should we remove this now as it is not used?

ryanskeith (dismissed):
The improvements all look sound.  

There is some code cleanup to do so whether it is done now or later.

conda_libmamba_solver/shards.py
dholth: Will remove.

travishathaway (changes requested):
This is looking good. My only problem are the two scripts you added the `tests` folder:

- `tests/httpx-fetch-all-shards.py`
- `tests/requests-fetch-all-shards.py`

They don't appear to be connected to anything else in the pull request. Can we please remove them? If you'd like to keep them in the repository, could we instead locate them at `dev/scripts`?

PR #697, opened 2025-09-03
sharded repodata model

### LINKED ISSUE #696: Add sharded repodata model, to unify sharded and non-sharded repodata handling
### Checklist

### What is the idea?

Add models to present all repodata as shards for a subsetting algorithm.

Add code to detect and fetch sharded repodata.

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

dholth:
Models to present repodata.json and repodata_shards.msgpack.zst in the same way; code to detect shards on the server.

Fix #696

### DIFF
--- a/conda_libmamba_solver/shards.py
+++ b/conda_libmamba_solver/shards.py
@@ -1,0 +1,338 @@
+"""
+Models for sharded repodata, and to make monolithic repodata look like sharded
+repodata.
+"""
+
+from __future__ import annotations
+
+import concurrent.futures
+import time
+from collections import defaultdict
+from typing import TYPE_CHECKING, TypedDict
+from urllib.parse import urljoin
+
+import conda.gateways.repodata as repodata
+import msgpack
+import zstandard
+from conda.base.context import context
+from conda.gateways.connection.session import get_session
+from conda.gateways.repodata import (
+    _add_http_value_to_dict,
+    conda_http_errors,
+)
+from conda.models.records import PackageRecord
+from requests import HTTPError
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from conda.core.subdir_data import SubdirData
+    from conda.gateways.repodata import (
+        RepodataCache,
+    )
+    from requests import Response
+
+
+Shard = TypedDict("Shard", {"packages": dict[str, dict], "packages.conda": dict[str, dict]})
+
+
+class RepodataInfo(TypedDict):  # noqa: F811
+    base_url: str  # where packages are stored
+    shards_base_url: str  # where shards are stored
+    subdir: str
+
+
+def maybe_unpack_record(record):
+    """
+    Convert bytes checksums to hex; leave unchanged if already str.
+    """
+    for hash_type in "sha256", "md5":
+        if hash_value := record.get(hash_type):
+            if isinstance(hash_value, bytes):
+                record[hash_type] = hash_value.hex()
+    return record
+
+
+def shard_mentioned_packages(shard: Shard):
+    """
+    Return all dependencies mentioned in a shard, including the shard's own
+    package name.
+
+    Includes virtual packages.
+    """
+    # XXX filter by package name for the possibility of a shard with multiple
+    # (small) packages
+    mentioned = set()
+    for package in (*shard["packages"].values(), *shard["packages.conda"].values()):
+        # to go faster, don't use PackageRecord, record.combined_depends, or
+        # MatchSpec
+        record = PackageRecord(**maybe_unpack_record(package))
+        mentioned.add(record.name)
+        mentioned.update(spec.name for spec in record.combined_depends)
+    return mentioned
+
+
+class ShardsIndex(TypedDict):
+    info: RepodataInfo
+    repodata_version: int
+    removed: list[str]
+    shards: dict[str, bytes]
+
+
+class ShardLike:
+    """
+    Present a "classic" repodata.json as per-package shards.
+    """
+
+    def __init__(self, repodata: dict, url: str = ""):
+        """
+        url: for debugging; not used by this class.
+        """
+        all_packages = {
+            "packages": repodata.pop("packages", {}),
+            "packages.conda": repodata.pop("packages.conda", {}),
+        }
+        self.repodata = repodata  # without packages, packages.conda
+        self.url = url
+
+        shards = defaultdict(lambda: {"packages": {}, "packages.conda": {}})
+
+        for package_group in all_packages:
+            for package, record in all_packages[package_group].items():
+                name = record["name"]
+                shards[name][package_group][package] = record
+
+        # defaultdict behavior no longer wanted
+        self.shards: dict[str, Shard] = dict(shards)  # type: ignore
+
+        # used to write out repodata subset
+        self.visited: dict[str, Shard | None] = {}
+
+    def __repr__(self):
+        return self.url.join(super().__repr__().split(maxsplit=1))
+
+    def __contains__(self, package):
+        return package in self.shards
+
+    def fetch_shard(self, package: str, session) -> Shard:
+        """
+        "Fetch" an individual shard.
+
+        Update self.visited with all not-None packages.
+
+        Raise KeyError if package is not in the index.
+        """
+        shard = self.shards[package]
+        self.visited[package] = shard
+        return shard
+
+    def fetch_shards(self, packages: Iterable[str], session) -> dict[str, Shard]:
+        """
+        Fetch multiple shards in one go.
+
+        Update self.visited with all not-None packages.
+        """
+        return {package: self.fetch_shard(package, session) for package in packages}
+
+    def build_repodata(self):
+        """
+        Return monolithic repodata including all visited shards.
+        """
+        repodata = self.repodata.copy()
+        repodata.update({"packages": {}, "packages.conda": {}})
+        for package, shard in self.visited.items():
+            if shard is None:
+                continue  # recorded visited but not available shards
+            for package_group in ("packages", "packages.conda"):
+                repodata[package_group].update(shard[package_group])
+        return repodata
+
+
+FETCHED_THIS_PROCESS = set()
+
+
+class Shards(ShardLike):
+    def __init__(self, shards_index: ShardsIndex, url: str):
+        """
+        Args:
+            shards_index: raw parsed msgpack dict
+            url: URL of repodata_shards.msgpack.zst
+        """
+        self.shards_index = shards_index
+        self.url = url
+
+        self.repodata = {k: v for k, v in self.shards_index.items() if k not in ("shards",)}
+
+        # used to write out repodata subset
+        self.visited: dict[str, Shard | None] = {}
+
+    def __contains__(self, package):
+        return package in self.packages_index
+
+    @property
+    def packages_index(self):
+        return self.shards_index["shards"]
+
+    def shard_url(self, package: str):
+        """
+        Return shard URL for a given package.
+
+        Raise KeyError if package is not in the index.
+        """
+        shard_name = f"{self.packages_index[package].hex()}.msgpack.zst"
+        # "Individual shards are stored under the URL <shards_base_url><sha256>.msgpack.zst"
+        return urljoin(self.url, f"{self.shards_index['info']['shards_base_url']}{shard_name}")
+
+    def fetch_shard(self, package: str, session) -> Shard:
+        """
+        Fetch an individual shard.
+
+        Raise KeyError if package is not in the index.
+        """
+
+        shard_url = self.shard_url(package)
+        raw_shard = session.get(shard_url).content
+        # ensure it is real msgpack+zstd before inserting into cache
+        shard: Shard = msgpack.loads(zstandard.decompress(raw_shard))  # type: ignore
+        self.visited[package] = shard
+        return shard
+
+    def fetch_shards(self, packages: Iterable[str], session) -> dict[str, Shard]:
+        """
+        Return mapping of *package names* to Shard for given packages.
+        """
+        result = {}
+
+        def fetch(s, url, package):
+            if url in FETCHED_THIS_PROCESS:
+                print("Already got", url)
+                raise RuntimeError("Can't fetch same url twice")
+            FETCHED_THIS_PROCESS.add(url)
+            b1 = time.time_ns()
+            data = s.get(url).content
+            e1 = time.time_ns()
+            print(f"Fetch took {(e1 - b1) / 1e9}s", package, url)
+            return (url, package, data)
+
+        packages = sorted(list(packages))
+        urls_packages = {self.shard_url(package): package for package in packages}
+
+        # beneficial to have thread pool larger than requests' default 10 max
+        # connections per session. There is "context.repodata_threads" but it's
+        # None in the REPL.
+        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+            futures = {
+                executor.submit(fetch, session, url, package): (url, package)
+                for url, package in urls_packages.items()
+                if package not in result
+            }
+            # May be inconvenient to cancel a large number of futures. Also, ctrl-C doesn't work reliably out of the box.
+            for future in concurrent.futures.as_completed(futures):
+                print(".", futures[future])
+                url, package, compressed_shard = future.result()
+                # XXX catch exception / 404 / whatever
+                result[package] = msgpack.loads(zstandard.decompress(compressed_shard))
+                try:
+                    package_names = [
+                        p["name"]
+                        for p in (
+                            *result[package]["packages"].values(),
+                            *result[package]["packages.conda"].values(),
+                        )
+                    ]
+                    print("expected", package, "actual", set(package_names))
+                except (AttributeError, KeyError):
+                    print("oops")
+
+        self.visited.update(result)
+
+        return result
+
+
+def repodata_shards(url, cache: RepodataCache) -> bytes:
+    """
+    Fetch shards index with cache.
+
+    Update cache state.
+
+    Return shards data, either newly fetched or from cache.
+    """
+    session = get_session(url)
+
+    state = cache.state
+    headers = {}
+    etag = state.etag
+    last_modified = state.mod
+    if etag:
+        headers["If-None-Match"] = str(etag)
+    if last_modified:
+        headers["If-Modified-Since"] = str(last_modified)
+    filename = "repodata_shards.msgpack.zst"
+
+    with conda_http_errors(url, filename):
+        timeout = (
+            context.remote_connect_timeout_secs,
+            context.remote_read_timeout_secs,
+        )
+        response: Response = session.get(
+            url, headers=headers, proxies=session.proxies, timeout=timeout
+        )
+        response.raise_for_status()
+        response_bytes = response.content
+
+    if response.status_code == 304:
+        # should we save cache-control to state here to put another n
+        # seconds on the "make a remote request" clock and/or touch cache
+        # mtime
+        return cache.cache_path_shards.read_bytes()
+
+    # We no longer add these tags to the large `resp.content` json
+    saved_fields = {repodata.URL_KEY: url}
+    _add_http_value_to_dict(response, "Etag", saved_fields, repodata.ETAG_KEY)
+    _add_http_value_to_dict(response, "Last-Modified", saved_fields, repodata.LAST_MODIFIED_KEY)
+    _add_http_value_to_dict(response, "Cache-Control", saved_fields, repodata.CACHE_CONTROL_KEY)
+
+    state.update(saved_fields)
+
+    # should we return the response and let caller save cache data to state?
+    return response_bytes
+
+
+def fetch_shards(sd: SubdirData) -> Shards | None:
+    """
+    Check a SubdirData's URL for shards.
+    Return shards index bytes from cache or network.
+    Return None if not found; caller should fetch normal repodata.
+    """
+
+    fetch = sd.repo_fetch
+    cache = fetch.repo_cache
+    # cache.load_state() will clear the file on JSONDecodeError but cache.load()
+    # will raise the exception
+    cache.load_state(binary=True)
+    cache_state = cache.state
+
+    if cache_state.should_check_format("shards"):
+        try:
+            # look for shards index
+            shards_index_url = f"{sd.url_w_subdir}/repodata_shards.msgpack.zst"
+            found = repodata_shards(shards_index_url, cache)
+            cache_state.set_has_format("shards", True)
+            # this will also set state["refresh_ns"] = time.time_ns(); we could
+            # call cache.refresh() if we got a 304 instead:
+            cache.save(found)
+
+            # basic parse (move into caller?)
+            shards_index: ShardsIndex = msgpack.loads(zstandard.decompress(found))  # type: ignore
+            shards = Shards(
+                shards_index,
+                shards_index_url,
+            )
+            return shards
+
+        except (HTTPError, repodata.RepodataIsEmpty):
+            # fetch repodata.json / repodata.json.zst instead
+            cache_state.set_has_format("shards", False)
+            cache.refresh(refresh_ns=1)  # expired but not falsy
+
+    return None
--- a/tests/test_shards.py
+++ b/tests/test_shards.py
@@ -1,0 +1,113 @@
+"""
+Test sharded repodata.
+"""
+
+from __future__ import annotations
+
+import json
+import random
+from pathlib import Path
+
+import pytest
+from conda.base.context import context, reset_context
+from conda.core.subdir_data import SubdirData
+from conda.gateways.connection.session import get_session
+from conda.models.channel import Channel
+
+from conda_libmamba_solver.shards import (
+    ShardLike,
+    fetch_shards,
+    shard_mentioned_packages,
+)
+
+HERE = Path(__file__).parent
+
+
+def package_names(shard):
+    """
+    All package names mentioned in a shard (should be a single package name)
+    """
+    return set(package["name"] for package in shard["packages"].values()) | set(
+        package["name"] for package in shard["packages.conda"].values()
+    )
+
+
+@pytest.fixture
+def conda_no_token(monkeypatch: pytest.MonkeyPatch):
+    """
+    Reset token to avoid being logged in. e.g. the testing channel doesn't understand them.
+    """
+    monkeypatch.setenv("CONDA_TOKEN", "")
+    reset_context()
+
+
+def test_shards(conda_no_token: None):
+    """
+    Test basic shard fetch etc.
+    """
+    channels = [
+        Channel.from_url(f"https://conda.anaconda.org/conda-forge-sharded/{subdir}")
+        for subdir in context.subdirs
+    ]
+
+    # Would eagerly download repodata.json.zst for all channels
+    # helper = LibMambaIndexHelper(
+    #     channels,
+    #     (),  # subdirs
+    #     "repodata.json",
+    #     installed,
+    #     (),  # pkgs_dirs to load packages locally when offline
+    #     in_state=in_state,
+    # )
+
+    # accessing "helper.repos" downloads repodata.json in the traditional way
+    # for all channels:
+    # print(helper.repos)
+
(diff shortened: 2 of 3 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
+# Copyright (C) 2022 Anaconda, Inc
+# Copyright (C) 2023 conda
```
dholth, line 2: Is "conda" an entity that can own a copyright?

jaimergp: 🤷

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+from conda.models.records import PackageRecord
+from requests import HTTPError
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from conda.core.subdir_data import SubdirData
+    from conda.gateways.repodata import (
+        RepodataCache,
+    )
```
jaimergp:
```suggestion
    from conda.gateways.repodata import RepodataCache
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+    from requests import Response
+
+
+Shard = TypedDict("Shard", {"packages": dict[str, dict], "packages.conda": dict[str, dict]})
+
+
+class RepodataInfo(TypedDict):  # noqa: F811
+    base_url: str  # where packages are stored
+    shards_base_url: str  # where shards are stored
+    subdir: str
```
jaimergp: Can we pick one style of defining the TypedDicts for consistency? I like the `class` version better.

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def maybe_unpack_record(record):
```
jaimergp:
```suggestion
def ensure_record_hash_is_hex(record: PackageRecord) -> PackageRecord:
```

I think this name is clearer.

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+    Convert bytes checksums to hex; leave unchanged if already str.
+    """
+    for hash_type in "sha256", "md5":
+        if hash_value := record.get(hash_type):
+            if isinstance(hash_value, bytes):
+                record[hash_type] = hash_value.hex()
+    return record
+
+
+def shard_mentioned_packages(shard: Shard):
```
jaimergp:
```suggestion
def shard_mentioned_packages(shard: Shard) -> set[str]:
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+    """
+    Return all dependencies mentioned in a shard, including the shard's own
```
jaimergp:
```suggestion
    Return all dependency names mentioned in a shard, including the shard's own
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+    for package in (*shard["packages"].values(), *shard["packages.conda"].values()):
+        # to go faster, don't use PackageRecord, record.combined_depends, or
+        # MatchSpec
+        record = PackageRecord(**maybe_unpack_record(package))
+        mentioned.add(record.name)
+        mentioned.update(spec.name for spec in record.combined_depends)
+    return mentioned
+
+
+class ShardsIndex(TypedDict):
```
jaimergp: Docstring? It would also be nice to move this next to the other typed dicts.

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+class ShardLike:
+    """
+    Present a "classic" repodata.json as per-package shards.
+    """
+
+    def __init__(self, repodata: dict, url: str = ""):
+        """
+        url: for debugging; not used by this class.
```
jaimergp: Should this be private then?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+        all_packages = {
+            "packages": repodata.pop("packages", {}),
+            "packages.conda": repodata.pop("packages.conda", {}),
+        }
+        self.repodata = repodata  # without packages, packages.conda
```
jaimergp: Maybe rename to "package_less_repodata"?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+                name = record["name"]
+                shards[name][package_group][package] = record
+
+        # defaultdict behavior no longer wanted
+        self.shards: dict[str, Shard] = dict(shards)  # type: ignore
+
+        # used to write out repodata subset
+        self.visited: dict[str, Shard | None] = {}
+
+    def __repr__(self):
```
jaimergp, line 132:
```suggestion
    def __repr__(self) -> str:
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        return self.url.join(super().__repr__().split(maxsplit=1))
+
+    def __contains__(self, package):
+        return package in self.shards
```
jaimergp:
```suggestion
    def __contains__(self, package_name: str) -> bool:
        return package_name in self.shards
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+
+        Update self.visited with all not-None packages.
+
+        Raise KeyError if package is not in the index.
+        """
+        shard = self.shards[package]
+        self.visited[package] = shard
+        return shard
+
+    def fetch_shards(self, packages: Iterable[str], session) -> dict[str, Shard]:
```
jaimergp: What's the session type?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+        Fetch multiple shards in one go.
+
+        Update self.visited with all not-None packages.
+        """
+        return {package: self.fetch_shard(package, session) for package in packages}
+
+    def build_repodata(self):
```
jaimergp:
```suggestion
    def build_repodata(self) -> dict:
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        self.shards_index = shards_index
+        self.url = url
+
+        self.repodata = {k: v for k, v in self.shards_index.items() if k not in ("shards",)}
+
+        # used to write out repodata subset
+        self.visited: dict[str, Shard | None] = {}
+
+    def __contains__(self, package):
+        return package in self.packages_index
```
jaimergp: Same as in parent class.

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+
+    @property
+    def packages_index(self):
```
jaimergp, line 202: Return type hint?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        return self.shards_index["shards"]
+
+    def shard_url(self, package: str):
```
jaimergp:
```suggestion
    def shard_url(self, package: str) -> str:
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+
+        shard_url = self.shard_url(package)
+        raw_shard = session.get(shard_url).content
+        # ensure it is real msgpack+zstd before inserting into cache
+        shard: Shard = msgpack.loads(zstandard.decompress(raw_shard))  # type: ignore
+        self.visited[package] = shard
+        return shard
+
+    def fetch_shards(self, packages: Iterable[str], session) -> dict[str, Shard]:
```
jaimergp: Type hint for session 🙏

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+        Return mapping of *package names* to Shard for given packages.
+        """
+        result = {}
+
+        def fetch(s, url, package):
+            if url in FETCHED_THIS_PROCESS:
```
jaimergp: might want to `global` this top-level name to avoid too many lookups?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+                print("Already got", url)
```
jaimergp: Remove or use `log.debug`?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+                raise RuntimeError("Can't fetch same url twice")
+            FETCHED_THIS_PROCESS.add(url)
+            b1 = time.time_ns()
+            data = s.get(url).content
+            e1 = time.time_ns()
+            print(f"Fetch took {(e1 - b1) / 1e9}s", package, url)
```
jaimergp: Remove or log.debug?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+            return (url, package, data)
+
+        packages = sorted(list(packages))
```
jaimergp, line 247: I don't think you need the list() call, do you?

```suggestion
        packages = sorted(packages)
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        urls_packages = {self.shard_url(package): package for package in packages}
+
+        # beneficial to have thread pool larger than requests' default 10 max
+        # connections per session. There is "context.repodata_threads" but it's
+        # None in the REPL.
```
jaimergp, line 258: Can you use something like this then?

```suggestion
        workers = context.repodata_threads or 10
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
```
jaimergp, line 259:
```suggestion
        with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,341 @@
...
+            futures = {
+                executor.submit(fetch, session, url, package): (url, package)
+                for url, package in urls_packages.items()
+                if package not in result
+            }
+            # May be inconvenient to cancel a large number of futures. Also, ctrl-C doesn't work reliably out of the box.
+            for future in concurrent.futures.as_completed(futures):
+                print(".", futures[future])
```
jaimergp: Remove debugging statement?

tests/test_shards.py
```diff
@@ -0,0 +1,116 @@
...
+HERE = Path(__file__).parent
+
+
+def package_names(shard):
+    """
+    All package names mentioned in a shard (should be a single package name)
+    """
+    return set(package["name"] for package in shard["packages"].values()) | set(
+        package["name"] for package in shard["packages.conda"].values()
+    )
```
jaimergp, line 46:
```suggestion
    return set(package["name"] for package in chain(shard["packages"].values(), shard["packages.conda"].values()))
```
Using `itertools.chain`

tests/test_shards.py
```diff
@@ -0,0 +1,116 @@
...
+        subdir_data.url_w_subdir
+    )  # XXX session could be different based on shards_base_url and different than the packages base_url
+
+    # download or fetch-from-cache a random set of shards
+
+    for package in random.choices([*found.packages_index.keys()], k=16):
+        shard = found.fetch_shard(package, session)
+
+        mentioned_in_shard = shard_mentioned_packages(shard)
+        print(package_names(shard), mentioned_in_shard)
```
jaimergp, line 126: Do you want to run any assertions here?

conda_libmamba_solver/shards.py
dholth: I don't know how to express "packages.conda" with a dot, in class style.

conda_libmamba_solver/shards.py
dholth: A hex upon you! I may have to define a PackageRecord TypedDict to implement those annotations.

conda_libmamba_solver/shards.py
dholth: It may be necessary to check conda to see whether the default has been lost. I was also having trouble with too many connections errors.

conda_libmamba_solver/shards.py
dholth: My editor knows it returns str anyway.

conda_libmamba_solver/shards.py
jezdez: I would fix this to read `conda community` since that refers back to the governance policy

conda_libmamba_solver/shards.py
dholth: Inherits from requests.Session

tests/test_shards.py
dholth: Maybe

conda_libmamba_solver/shards.py
dholth: (is part of the pre-commit automation)

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,356 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from typing import NotRequired
+
+    from conda.core.subdir_data import SubdirData
+    from conda.gateways.repodata import RepodataCache
+    from requests import Response
+
+
+class PackageRecordDict(TypedDict):
```
dholth: Not "closed=True", can contain other keys

conda_libmamba_solver/shards.py
jezdez: Ah, seems like I knew this at some point, https://github.com/conda/conda-libmamba-solver/pull/167. This must be fine since IIRC we talked to legal at the time.

conda_libmamba_solver/shards.py
dholth: My editor also knows the return type already?

tests/test_shards.py
dholth: IMO set union is probably as efficient

conda_libmamba_solver/shards.py
dholth: Added in the combined PR version https://github.com/conda/conda-libmamba-solver/pull/686/files#diff-09617837bcdab07fd05c2faebb3585caab01a13ee029e1af4be262936ba0b4b7R62

conda_libmamba_solver/shards.py
dholth: https://github.com/conda/conda-libmamba-solver/pull/686/files#diff-09617837bcdab07fd05c2faebb3585caab01a13ee029e1af4be262936ba0b4b7R116 "repodata_no_packages"

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,389 @@
...
+from __future__ import annotations
+
+import concurrent.futures
+import logging
+import time
+from collections import defaultdict
+from typing import TYPE_CHECKING, TypedDict
+from urllib.parse import urljoin
+
+import conda.gateways.repodata as repodata
```
jezdez: Minor code smell: importing the module as a global module variable `repodata` means it overlaps a local variable with the same name in some of the methods below.

On [line 23-26](https://github.com/conda/conda-libmamba-solver/pull/697/files#diff-09617837bcdab07fd05c2faebb3585caab01a13ee029e1af4be262936ba0b4b7R23-R26) we import a few items from that module already, can we do that with the other items as well? I think it's just some constants and an exception class.

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,389 @@
...
+    """
+    Basic package attributes that this module cares about.
+    """
+
+    name: str
+    sha256: NotRequired[str | bytes]
+    md5: NotRequired[str | bytes]
+
+
+# in this style because "packages.conda" is not a Python identifier
```
jezdez: 👍🏻

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,389 @@
...
+    repodata_version: int
+    removed: list[str]
+    shards: dict[str, bytes]
+
+
+def ensure_hex_hash(record: PackageRecordDict):
+    """
+    Convert bytes checksums to hex; leave unchanged if already str.
+    """
+    for hash_type in "sha256", "md5":
```
jezdez, line 75:
```suggestion
    for hash_type in ("sha256", "md5"):
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,389 @@
...
+    for package in (*shard["packages"].values(), *shard["packages.conda"].values()):
+        # to go faster, don't use PackageRecord, record.combined_depends, or
+        # MatchSpec
+        record = PackageRecord(**ensure_hex_hash(package))
+        mentioned.add(record.name)
+        mentioned.update(spec.name for spec in record.combined_depends)
+    return mentioned
+
+
+class ShardLike:
```
jezdez, line 101: 🦈

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,389 @@
...
+        if package in self.visited:
+            log.debug("Revisit %s %s", self.url, package)
+            shard_or_none = self.visited[package]
+            if shard_or_none is None:
+                raise KeyError(package)
+            return shard_or_none
+
+        shard_url = self.shard_url(package)
+
+        raw_shard = self.session.get(shard_url).content
```
jezdez: We might want to do some defensive HTTP error handling here

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,389 @@
...
+        def fetch(s, url, package):
+            if url in FETCHED_THIS_PROCESS:
+                log.debug("Already got %s", url)
+                raise RuntimeError("Can't fetch same url twice")
+            FETCHED_THIS_PROCESS.add(url)
+            b1 = time.time_ns()
+            data = s.get(url).content
+            e1 = time.time_ns()
+            log.debug(f"Fetch took {(e1 - b1) / 1e9}s", package, url)
+            return (url, package, data)
```
```suggestion
        def fetch(session, url, package):
            if url in FETCHED_THIS_PROCESS:
                log.debug("Already got %s", url)
                raise RuntimeError("Can't fetch same url twice")
            FETCHED_THIS_PROCESS.add(url)
            pre_timestamp = time.time_ns()
            data = session.get(url).content
            post_timestamp = time.time_ns()
            request_time = (post_timestamp - pre_timestamp) / 1e9
            log.debug(f"Fetch took {request_time}s", package, url)
            return (url, package, data)
```

Also, `session.get` calls need some form of error handling, retries etc?

conda_libmamba_solver/shards.py
jezdez: Or are you doing this in a different layer elsewhere?

conda_libmamba_solver/shards.py
dholth: Changed

conda_libmamba_solver/shards.py
dholth: I don't get this suggestion.

conda_libmamba_solver/shards.py
dholth: Shard week

conda_libmamba_solver/shards.py
dholth: Will have to refactor the networking into another module...

dholth: [CondaSession configures retries](https://github.com/conda/conda/blob/main/conda/gateways/connection/session.py#L209)

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,379 @@
...
+
+        def fetch(s, url, package):
+            if url in FETCHED_THIS_PROCESS:
+                log.debug("Already got %s", url)
+                raise RuntimeError("Can't fetch same url twice")
+            FETCHED_THIS_PROCESS.add(url)
+            b1 = time.time_ns()
+            data = s.get(url).content
+            e1 = time.time_ns()
+            log.debug(f"Fetch took {(e1 - b1) / 1e9}s", package, url)
```
jaimergp: These arguments are not being used, are they?

dholth: It was a print statement, needs to be updated

isuruf:
Is the plan that these models get transferred to some other package? They don't seem `conda-libmamba-solver` specific.

dholth:
How can we move this forward?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,382 @@
...
+        repodata.pop("packages.conda")
+        self.repodata_no_packages = repodata  # without packages, packages.conda
+        self.url = url
+
+        shards = defaultdict(lambda: {"packages": {}, "packages.conda": {}})
+
+        for package_group in all_packages:
+            for package, record in all_packages[package_group].items():
+                name = record["name"]
+                shards[name][package_group][package] = record
```
jaimergp:
```suggestion
        for group_name, group in all_packages.items():
            for package, record in group.items():
                name = record["name"]
                shards[name][group_name][package] = record
```

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,382 @@
...
+    def __init__(self, repodata: RepodataDict, url: str = ""):
+        """
+        url: affects the repr but not the functionality of this class.
+        """
+        all_packages = {
+            "packages": repodata["packages"],
+            "packages.conda": repodata["packages.conda"],
+        }
+        repodata.pop("packages")
+        repodata.pop("packages.conda")
```
jaimergp, line 115:
```suggestion
        all_packages = {
            "packages": repodata.pop("packages"),
            "packages.conda": repodata.pop("packages.conda"),
        }
```

dholth: I think it was written this way for the type checker, but we should show the type checker who's boss and write the code more nicely. Do we accept KeyError on bad inputs (missing packages, packages.conda)?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,382 @@
...
+
+    @property
+    def packages_index(self):
+        return self.shards_index["shards"]
+
+    @property
+    def base_url(self) -> str:
+        """
+        Return self.url joined with shards_base_url.
+        Note shards_base_url can be a relative or an absolute url.
```
jaimergp, line 209: In this case the result of `urljoin` is to return the rightmost absolute URL, right?

dholth: Yes that's the part that is more special than "strip rightmost / and string join" e.g.

tests/test_shards.py
```diff
@@ -2,25 +2,159 @@
...
+    )
+
+
+def test_fetch_shards_error(http_server_shards):
+    channel = Channel.from_url(f"{http_server_shards}/noarch")
+    subdir_data = SubdirData(channel)
+    found = fetch_shards(subdir_data)
+    assert found
+
+    with pytest.raises(zstandard.ZstdError):
```
jaimergp, line 92: Is there a way to raise a more precise exception? Although I guess the user will never see this, would they? What happens if they run something like `conda create -n something -c sharded-channel does-not-exist`?

tests/test_shards.py
```diff
@@ -2,25 +2,159 @@
...
+        # XXX we tried to decompress a 404 response
+        found.fetch_shard("fake_package")
+
+    malo = found.fetch_shard("malformed")
```
jaimergp, line 96: 🇪🇸

conda_libmamba_solver/shards.py
jaimergp: You can `.pop(X, {})` in those cases and swallow the KeyError, right?

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,382 @@
...
+    for package in (*shard["packages"].values(), *shard["packages.conda"].values()):
+        # to go faster, don't use PackageRecord, record.combined_depends, or
+        # MatchSpec
+        record = PackageRecord(**ensure_hex_hash(package))
+        mentioned.add(record.name)
+        mentioned.update(spec.name for spec in record.combined_depends)
+    return mentioned
+
+
+class ShardLike:
```
jaimergp, line 101: I have a question about the purpose of this class at https://github.com/conda/conda-libmamba-solver/pull/686/files#r2356133965. Do we really need it?

conda_libmamba_solver/shards.py
dholth: The type checker would prefer I replace the keys with {} which is still dict[str, dict] we'll see.

tests/test_shards.py
dholth: If the package name isn't in the shards index, that's not an error.
If the shards index can't be loaded for any reason, including compression or data structure errors, we should fall back to monolithic and remember that for 7 days.
This checks if the shard listed in the index is a 404 or isn't zstandard-compressed.

tests/test_shards.py
```diff
@@ -2,25 +2,159 @@
...
+
+
+def test_fetch_shards_error(http_server_shards):
+    channel = Channel.from_url(f"{http_server_shards}/noarch")
+    subdir_data = SubdirData(channel)
+    found = fetch_shards(subdir_data)
+    assert found
+
+    with pytest.raises(zstandard.ZstdError):
+        # XXX we tried to decompress a 404 response
```
dholth:
```suggestion
        # XXX this is currently trying to decompress the server's 404 response, which should be a `requests.Response.raise_for_status()`
```

conda_libmamba_solver/shards.py
dholth: Yes, we do need to present "every release of a given package name in a channel" in a uniform way for the traversal algorithm, and this implementation should be pretty quick.
We could choose to write the subset only for a real shards channel and point to the cached json for shardlike channels.
Ideally the sharded channel is "the big one" and the users' other channels are not a bottleneck.

conda_libmamba_solver/shards.py
```diff
@@ -0,0 +1,382 @@
...
+import msgpack
+import zstandard
+from conda.base.context import context
+from conda.gateways.connection.session import get_session
+from conda.gateways.repodata import (
+    _add_http_value_to_dict,
+    conda_http_errors,
+)
+from conda.models.records import PackageRecord
+from requests import HTTPError
```
travishathaway, line 29: We should update our project dependencies (e.g. `recipe/meta.yaml` and `pyproject.toml`) to reflect the fact that that `msgpack`, `zstandard` and `requests` are now primary dependencies and required for this library.

Let's not rely on the fact that these are included as transitive dependencies from other dependencies (e.g. `conda` and `conda-index`). This is especially the case for `msgpack` because a default Miniconda installation will not include `conda-index` and therefore `msgpack`.

The only reason the tests are passing now is because `conda-index` is listed as a testing requirement.

dholth: Do it!

conda_libmamba_solver/shards.py
jaimergp: > We could choose to write the subset only for a real shards channel and point to the cached json for shardlike channels.

I think this is desirable. I'll leave it up to you whether you want to do it here or in another PR.

jaimergp: Also, from our conversation about the review we could add your explanation somewhere (comment or docstring) about _why_ we need this. IIUC, this is supporting flexible channel priority when sharded and non-sharded channels are mixed in the same solve.

dholth: You're right, we don't explain what we are doing, we just jump into the algorithm.

travishathaway (approved):

conda_libmamba_solver/shards.py
jaimergp: @dholth, this comment is the only thing holding my approval. We can either do it here on in the next PR. Your call. I'll approve for now in good faith! Thanks!

Issue #709, opened 2025-07-21
Discovery: Implement Performance Testing Framework for Sharded Repodata


george-lorch:
**What is the idea?**
Create comprehensive performance tests to measure and validate the improvements provided by the shard-based approach.

**Why is this needed?**
Performance improvements need to be measurable and verifiable. Testing helps identify bottlenecks and ensures that the new system actually provides the expected benefits over the current approach.

**What should happen?**

- [ ] Research potential ways to measure performance of sharded repodata (leave comments in this issue)
- [ ] Come up up with a recommendation for how to generate a report similar to [prefix's sharded repodata blog post](https://prefix.dev/blog/sharded_repodata)

**Additional Context**
Include benchmarks for both cold cache and warm cache scenarios. Consider testing across different network conditions and repository configurations.

This is an example of how we are currently tracking this performance improvements:

- https://github.com/dholth/conda-libmamba-solver/blob/sharded-experiments/tests/test_shards.py#L368

travishathaway:
I've started this document as way to organize my thoughts around how I'd like to complete this issue:

- https://hackmd.io/dKEGO6ybRWWhinibwrgDBQ

travishathaway:
Another thing worth mentioning is that the sharded repodata code itself isn't complete at this time. If something hasn't been merged into the main branch, it makes it impossible to write benchmarking tests for. This is what led me to pursuing a more generic benchmarking tool.

danyeaw:
Worked completed for discovery phase.

Issue #705, opened 2025-09-17
Support Sharded Repodata 2025Q4


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open issues and couldn't find a duplicate

### What?

This is the overarching epic for all work related to adding sharded repodata support to conda-libmamba-solver.

### Why?

We are completing this work to support [CEP-16: Sharded Repodata](https://conda.org/learn/ceps/cep-0016). The ultimate goal is to speed up retrieval of repodata records from remote channels, especially large channels like "conda-forge".

### User impact

- Faster download times of repodata leading to an overall faster experience for conda

### Goals

- [ ] Faster download times of repodata
- [ ] Less demand on channel servers like anaconda.org

#### Core functionality

- [x] https://github.com/conda/conda-libmamba-solver/issues/706
- [x] https://github.com/conda/conda-libmamba-solver/issues/707
- [x] https://github.com/conda/conda-libmamba-solver/issues/684
- [x] https://github.com/conda/conda-libmamba-solver/issues/710
- [x] https://github.com/conda/conda-libmamba-solver/issues/696
- [x] https://github.com/conda/conda-libmamba-solver/issues/711
- [x] https://github.com/conda/conda-libmamba-solver/issues/712
- [x] https://github.com/conda/conda-libmamba-solver/issues/696
- [x] https://github.com/conda/conda-libmamba-solver/issues/715
- [x] https://github.com/conda/conda-libmamba-solver/issues/725

#### Testing and performance

- [x] https://github.com/conda/conda-libmamba-solver/issues/708
- [x] https://github.com/conda/conda-libmamba-solver/issues/709
- [x] https://github.com/conda/conda-libmamba-solver/issues/721

#### Miscellaneous
- [x] https://github.com/conda/conda/issues/15233

PR #714, opened 2025-09-22
Create a script to profile conda processes

travishathaway:
### Description

Closes: #709

I want to state the obvious by saying this falls well short of creating a, "Performance Testing Framework". Instead, my goal with this pull request is to create a script that we can manually run to get a set basic statistics regarding performance improvements for our sharded repodata project (and potentially beyond).

I do not think this project is the best place for this pull request, but it is closest to the project that is currently being worked on that needs profiling (#705).

For the review, I just need to know whether there are any obvious errors or problematic assumptions the code is making regarding the profiling it does. I don't actually think we will merge this code into this repository.

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-libmamba-solver/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~

### DIFF
--- a/profiling/Dockerfile
+++ b/profiling/Dockerfile
@@ -1,0 +1,6 @@
+FROM continuumio/miniconda3:latest
+
+RUN conda install --yes psutil
+COPY profile.py /root/
+
+CMD ["python", "/root/profile.py"]
--- a/profiling/profile.py
+++ b/profiling/profile.py
@@ -1,0 +1,83 @@
+"""
+Profiles conda processes using psutil plus a couple of built-in modules
+
+NOTICE: this currently only works on Unix-like systems (macOS and Linux)
+"""
+import resource
+import subprocess
+import sys
+import time
+
+import psutil
+
+#: Value used to convert bytes to megabytes
+BYTES_TO_MB = 1024 * 1024
+
+#: Value used to convert kilobytes to megabytes
+KILOBYTES_TO_MB = 1024
+
+
+def main():
+    """
+    Entry point for the profiling script
+    """
+
+    # Gather the network and time statistics before
+    net_io_stats_before = psutil.net_io_counters()
+    time_before = time.time()
+
+    command = [
+        "python", "-m",
+        # conda command
+        "conda", "create",
+        # settings
+        "--name", "profile-env",
+        "--channel", "conda-forge",
+        "--override-channels",
+        "--yes",
+        "--quiet",
+        # packages
+        "python",
+    ]
+    sub = subprocess.Popen(command)
+    sub.wait()
+
+    rusage = resource.getrusage(resource.RUSAGE_CHILDREN)
+
+    # Gather statistics after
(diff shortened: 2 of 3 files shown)


profiling/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM continuumio/miniconda3:latest
```
dholth, line 1: Weird that we never got the current company name into this docker URL

profiling/profile.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+import resource
+import subprocess
+import sys
+import time
+
+import psutil
+
+#: Value used to convert bytes to megabytes
+BYTES_TO_MB = 1024 * 1024
```
dholth, line 14: 2**20 or 1<<20

profiling/profile.py
```diff
@@ -0,0 +1,83 @@
...
+
+#: Value used to convert kilobytes to megabytes
+KILOBYTES_TO_MB = 1024
```
dholth, line 17: 2**10 or 1<<10

profiling/profile.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def main():
+    """
+    Entry point for the profiling script
+    """
+
+    # Gather the network and time statistics before
+    net_io_stats_before = psutil.net_io_counters()
```
dholth, line 26: Neat technique

profiling/profile.py
```diff
@@ -0,0 +1,83 @@
...
+    max_memory_usage = round(rusage.ru_maxrss / max_mem_convert, ndigits=2)
+
+    command_str = " ".join(command)
+
+    print()
+
+    print("Statistics for:", command_str)
+
+    # Network
+    print(f"Received: {bytes_recv}MB")
```
dholth, line 70:
```suggestion
    print(f"Received: {bytes_recv:.2f}MB")
```
Instead of rounding in the `round()` function

profiling/profile.py
```diff
@@ -0,0 +1,83 @@
...
+        "python",
+    ]
+    sub = subprocess.Popen(command)
+    sub.wait()
+
+    rusage = resource.getrusage(resource.RUSAGE_CHILDREN)
+
+    # Gather statistics after
+    net_io_stats_after = psutil.net_io_counters()
+    time_after = time.time()
```
dholth, line 49:
```suggestion
    time_after = time.monotonic_ns()
```
    
Gives the time in nanoseconds, avoiding floating point errors or system clock updates.

> Return the value (in fractional seconds) of a monotonic clock, i.e. a clock that cannot go backwards. The clock is not affected by system clock updates. The reference point of the returned value is undefined, so that only the difference between the results of two calls is valid.

travishathaway:
Thanks @dholth!

PR #731, opened 2025-10-10
Use (channel, package) shards to improve sharded subset

### LINKED ISSUE #730: Refactor shard traversal to use (channel, package name) instead of just (package name)
### Checklist

### What is the idea?

By using (channel, package name) as nodes in shard traversal, the rest of the code is simplified as it no longer has to broadcast over all channels. This may help with optimizations e.g. traverse all monolithic repodata first, then sqlite3 cached repodata; be smarter about connection pools, ...

dholth:
### Description

Use (channel, package) to find shards instead of broadcasting package names x channels in several places.

Fix #730 

### Checklist - did you ...


conda_libmamba_solver/shards.py
```diff
@@ -487,65 +487,47 @@ def fetch_channels(channels):
...
-                # urljoin consistent with shards which end with
-                # /repodata_shards.msgpack.zst
-                url = f"{channel_url}/repodata.json"
-                found = ShardLike(repodata_json, url)
+    # The parallel version may reorder channels, does this matter?
+
+    # beneficial to have thread pool larger than requests' default 10 max
+    # connections per session. There is "context.repodata_threads" but it's
+    # None in the REPL. For channels, we would usually be limited by the
+    # number of channels but not for individual shards elsewhere in this code.
```
dholth: It may not actually be useful to exceed the connection pool size.
One way of getting the pool size, which is possibly always requests' default for conda:
`get_session(url).get_adapter('https://').poolmanager.connection_pool_kw['maxsize']`

conda_libmamba_solver/shards.py
```diff
@@ -438,6 +456,8 @@ def batch_retrieve_from_cache(sharded: list[Shards], packages: list[str]):
     sharded = [shardlike for shardlike in sharded if isinstance(shardlike, Shards)]
 
     wanted = []
+    # XXX update batch_retrieve_from_cache to work with (Shards, package name)
+    # tuples instead of broadcasting across shards itself.
```
dholth, line 484: batch_retrieve_from_cache will change in the multithreaded refactor, no reason to change it here

conda_libmamba_solver/shards.py
```diff
@@ -46,6 +45,33 @@
...
+    Is this shared between all sessions? Or do we get a different pool for a
+    different get_session(url)?
+
+    Other adapters (file://, s3://) used in conda would have different
+    concurrency behavior;  we are not prepared to have separate threadpools per
+    connection type.
+    """
+    if context.repodata_threads is not None:
+        return context.repodata_threads
+    session = get_session("https://repo.anaconda.com/pkgs/main")
```
jaimergp: Is this harcoded channel necessary or shall we pull from `context.channels`?

conda_libmamba_solver/shards.py
```diff
@@ -46,6 +45,33 @@
...
+    adapter = session.get_adapter("https://")
+    if poolmanager := getattr(adapter, "poolmanager"):
+        try:
+            return int(poolmanager.connection_pool_kw["maxsize"])
+        except (KeyError, ValueError, AttributeError):
```
jaimergp: Should probably catch `TypeError` too to avoid exceptions from things like `int(None)`.

conda_libmamba_solver/shards.py
dholth: We could look for the first context.channels that is an https:// or return the default value if none of those channels are. Ideally it would be something that is re-used in the get_session() functools.cache(). Any https:// URL gives us the pool number of threads value that we are looking for.

conda_libmamba_solver/shards_subset.py
```diff
@@ -59,12 +59,31 @@
 class Node:
     distance: int = sys.maxsize
     package: str = ""
+    channel: str = ""
     visited: bool = False
 
+    def to_id(self):
```
jaimergp:
```suggestion
    def to_id(self) -> NodeId:
```

conda_libmamba_solver/shards_subset.py
```diff
@@ -59,12 +59,31 @@
...
+        return NodeId(self.package, self.channel)
+
+    def in_shard(self, shardlike: ShardLike):
```
jaimergp:
```suggestion
    def in_shard(self, shardlike: ShardLike) -> bool:
```

conda_libmamba_solver/shards.py
dholth:
```suggestion
        except (KeyError, ValueError, AttributeError, TypeError):
```

conda_libmamba_solver/shards_subset.py
```diff
@@ -76,32 +95,25 @@ def neighbors(self, node: Node):
...
 
-                        # yield self.nodes[package]
+                if node_id not in self.nodes:
+                    self.nodes[node_id] = Node(node.distance + 1, package, shardlike.url)
+                    yield self.nodes[node_id]
 
-                    discovered.add(package)  # also doesn't make it faster
+                    if package not in discovered:
+                        # now this is per package name, not per (name, channel) tuple
+                        log.debug(f"{json.dumps(node.package)} -> {json.dumps(package)};")
```
jaimergp:
```suggestion
                        log.debug("%s -> %s;", json.dumps(node.package, json.dumps(package))
```

tests/test_solver_differences.py
```diff
@@ -144,8 +144,17 @@ def test_gpu_cpu_mutexes():
...
+    # cudatoolkit is not installed when cpuonly is specified.
+    found = 0
+    target_pkgs = ("pytorch", "pyg")
+    for pkg in data["actions"]["LINK"]:
+        if pkg["name"] in target_pkgs:
+            found += 1
+            assert "cpu" in pkg["build_string"]
+        elif pkg["name"] == "cudatoolkit":
+            raise AssertionError("CUDA shouldn't be installed due to 'cpuonly'")
+    assert found == len(target_pkgs)
```
jaimergp, line 158: Addressed in https://github.com/conda/conda-libmamba-solver/pull/741, please merge/rebase main :)

conda_libmamba_solver/shards.py
jaimergp: So we just need a dummy session to inspect, correct? Isn't it enough with a blank `CondaSession()`?

jaimergp: I see you marked this for automerge, could you look into this @dholth? Does it work with `CondaSession()`?

dholth: Changed to CondaSession(), tests still pass.

conda_libmamba_solver/shards.py
```diff
@@ -46,6 +45,37 @@
 
     from .shards_typing import PackageRecordDict, ShardDict
 
+SHARDS_CONNECTIONS_DEFAULT = 10
+ZSTD_MAX_SHARD_SIZE = 2**20 * 16  # maximum size necessary when compresed data has no size header
```
ryanskeith: It might be nice to have the approximate number of ~16.8 Million  listed for comparison.

conda_libmamba_solver/shards_subset.py
```diff
@@ -59,12 +59,31 @@
...
     visited: bool = False
 
+    def to_id(self) -> NodeId:
+        return NodeId(self.package, self.channel)
+
+    def in_shard(self, shardlike: ShardLike) -> bool:
+        return self.channel == shardlike.url
+
+
+@dataclass(order=True, eq=True, frozen=True)
```
ryanskeith, line 72: I know this is functionally the same-ish, but would using the NameTuple offer any benefit?  I think their memory footprint is a bit smaller.

ryanskeith (dismissed):
Looks good to me.  I had a couple comments that may be worth addressing.

conda_libmamba_solver/shards.py
dholth: Maybe, but everyone should know that `2**20` is one megabyte. Oh, this number is also `2**24`

conda_libmamba_solver/shards_subset.py
dholth: Here we want the class to have a method and we don't want it to be a tuple.

tests/test_shards.py
```diff
@@ -644,3 +648,22 @@ def test_batch_retrieve_from_cache(prepare_shards_test: None):
...
+
+def test_shards_connections(monkeypatch):
+    """
+    Test _shards_connections() and execute all its code.
+    """
+    assert context.repodata_threads is None
+    assert _shards_connections() == 10  # requests' default
+
+    poolmanager = (
+        get_session("https://repo.anaconda.com/pkgs/main").get_adapter("https://").poolmanager
```
jaimergp:
```suggestion
        CondaSession().get_adapter("https://").poolmanager
```

Here as well?

jaimergp (dismissed):

jaimergp (approved):

PR #741, opened 2025-10-15
CUDA/CPU mutex handling for PyTorch and PyTorch Geometric now works; update tests and docs

### LINKED ISSUE #131: Solver installs pytorch CUDA variants when CPU ones are requested via `cpuonly`
_Originally posted by @agitter in https://github.com/conda/conda-libmamba-solver/issues/115#issuecomment-1399040871_

---

I experienced an issue that is somewhat related but not identical.  If it is sufficiently different, I'll open a new issue.

In my case, I was installing the same PyTorch environment with `classic` and `libmamba` solvers:
```
$ conda env create -f pyg-minimal.yml --experimental-solver classic
```
```
$ conda env create -f pyg-minimal.yml --experimental-solver libmamba
```
The `classic` solver correctly installed the CPU versions but the `libmamba` solver incorrectly installed `cudatoolkit` and the CUDA versions.

`pyg-minimal.yml`:
```
name: pyg
channels:
  - conda-forge
  - pyg
  - pytorch
dependencies:
  - cpuonly
  - pyg=2.1.0
  - python=3.9
  - pytorch::pytorch=1.12
```

agriyakhetarpal:
### Description

See recent discussion in #131. This PR updates the `test_gpu_cpu_mutexes` test and related documentation about differences between the classic and `libmamba` solvers to reflect that issue #131 has been resolved. The `libmamba` solver now correctly respects the `cpuonly` mutex and no longer installs CUDA packages when CPU-only variants are requested, matching the behaviour of the classic solver.

This means that under-specified libraries, like older pyg packages that relied on transitive mutexes for CPU variants, as opposed to well-packaged libraries, can now prevent CUDA variants from being selected.

I have moved the previous difference to a "Historical" section and rewritten the paragraph in the past tense where applicable.

Closes #131

### Questions

Do we need to add a minimum version bound for `libmamba` or `libsolv` somewhere?

### Checklist - did you ...

### DIFF
--- a/docs/user-guide/libmamba-vs-classic.md
+++ b/docs/user-guide/libmamba-vs-classic.md
@@ -311,7 +311,7 @@ thus compatible with _any_ Python version.
 
 This section documents solver behavior differences that existed in the past but have since been
 resolved. These are kept for historical reference, and to help users understand the evolution of
-the `conda-libmamba-solver`.
+`conda-libmamba-solver`.
 
 ### cudatoolkit present in a `cpuonly` environment
 


agriyakhetarpal:
`test_gpu_cpu_mutexes` passing from a PR that I opened from this branch on my fork: https://github.com/agriyakhetarpal/conda-libmamba-solver/actions/runs/18514437599/job/52761979559?pr=2#step:14:142

tests/test_solver_differences.py
```diff
@@ -157,7 +166,7 @@ def test_gpu_cpu_mutexes():
         env=env,
     )
     data = json.loads(p.stdout)
-    # This should not happen, but it does. See docstring.
```
agriyakhetarpal, line 160: This test should pass and did pass because newer PyTorch versions were always properly packaged. So this comment, "This should not happen, but it does", did not make sense here.

I think what happened is that:
- the comment was copied from the previous test section above (where the bug _did_ occur)
- but this third test section was actually checking the correct behaviour (i.e., no `cudatoolkit` with newer versions)
- the comment was never updated to match what's happening

I've changed the comment to say "With more recent pytorch versions, this works correctly, as there is no cudatoolkit.", as they properly declared the muxet.

docs/user-guide/libmamba-vs-classic.md
```diff
@@ -307,17 +307,26 @@ the expense of an older `python`, and `conda-libmamba-solver` will prefer having
 even if it means going all the way down to `pydantic 0.18.2` (which was packaged as `noarch`) and
 thus compatible with _any_ Python version.
 
+## Historical issues related to solver differences (resolved)
+
+This section documents solver behavior differences that existed in the past but have since been
+resolved. These are kept for historical reference, and to help users understand the evolution of
+the `conda-libmamba-solver`.
```
jaimergp:
```suggestion
`conda-libmamba-solver`.
```

tests/test_solver_differences.py
```diff
@@ -144,8 +144,17 @@ def test_gpu_cpu_mutexes():
...
-    assert next(pkg for pkg in data["actions"]["LINK"] if pkg["name"] == "cudatoolkit")
+    # After recent libsolv/libmamba updates, this now works correctly.
```
jaimergp, line 147: Can you try to pinpoint in which version the fix was introduced? 🙏

agriyakhetarpal: I had the same question (as I added in the PR description) – I'll try to scope out the difference with what I can see from CI runs.

dholth: or "pytorch repackaging" etc?

PR #743, opened 2025-10-15
Improved threading in shards traversal

### LINKED ISSUE #736: Pipelined Shard Resolution
### What is the idea?

Sharded repodata runs a reachability algorithm on each dependency, following all dependency names in each shard to find new shards starting from the installed and requested packages. This generous subset of "any package that could possibly be required" is sent to the solver, and is much smaller than the complete repodata.

We can estimate the time to fetch shards as `(number of shards / parallelism) * request latency`. This is a heuristic; some shards are more distant in the dependency graph and must be discovered by visiting the initial shards.

The less-parallel algorithm visits each shard, waits for each dependent shard to be fetched, and visits those nodes in some order. It also tries to batch-fetch any missing shards in the queue from cache, and then the network, but it is not doing other work while those shards are being fetched. After the fetch step completes, the number of http requests in-flight drops to zero.

This is bad for performance. Suppose we have 10 http requests in-flight, we receive 9 of them, but one is especially slow; we have to wait for that slow request before firing off new http requests. The latency term in our estimate becomes the slowest request out of each batch, and multiplies proportional to the number of shards.

Instead, we propose to visit shards "as soon as they are loaded".

We can implement this with parallel tasks that communicate with queues.

```mermaid
sequenceDiagram
    loop
        Main ->> Main: "Fetch" in-memory shard
        Main ->> Cache: Fetch shard
        Cache ->> Network: Cache miss
        Cache ->> Main: Cache hit
        Network ->> Main: Network result
        Main ->> Main: Find new (channel, package) from shard data
    end
```

The main thread receives shards from a shard-out queue. When it sees a shard, it finds all shard id's in that shard that have not already been visited, sending them to the shard-in queue. Any shard id sent to the shard-in queue will eventually show up as shard data on the shard-out queue.

The cache thread loads shards from sqlite3. Hits go into the shard-out queue, and misses go into the cache-miss queue which has the same type signature as the shard-in queue.

The network thread loads shards from URLs, sending responses to the shard-out queue as soon as they are received, and putting those responses in the cache.

In this way, there is no delay between receiving a shard from the cache or network, and finding new shard id's to fetch. So we begin waiting for cache and network responses as soon as possible, instead of after the slowest shard in a batch of "n" shards has been received.

This implementation also decouples the main thread from the cache and network implementation. It only knows that shard id's sent to the shard-in queue eventually appear on the shard-out queue. We also have the opportunity to [replace the network implementation with http/2](https://github.com/conda/conda-libmamba-solver/pull/774) without affecting the main thread.

The desired outcome is reduced repodata fetch time especially for cold-cache scenarios, better CPU/network utilization, and improved cache efficiency.

- [x] Benchmark new approach to prove new method is faster by adding new entries (e.g. `shortest_threaded_pipeline`) in these tests:
  - [x] `tests/test_shards_subset.py::test_traversal_algorithm_benchmarks`
  - [x] `tests/test_shards_subset.py::test_traversal_algorithms_match`

### Additional Context

Main changes will be in `shards.py`, `shards_cache.py`, `shards_subset.py`, and `solver.py`.

dholth:
### Description

Workers for sqlite3 cache retrieval, network retrieval.

While shard traversal is happening, workers continuously empty queue of shards to fetch; try to fetch those shards in as large a batch as is available; and then dispatch negative results to the network worker or positive results to the queue of fetched shards.

The main thread finds dependencies for packages as soon as we have the repodata for those packages. Any missing URLs are dispatched to worker threads.

Fix #736

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -435,7 +435,7 @@ jobs:
         run: conda list --show-channel-urls
 
       - name: Run Benchmarks
-        uses: CodSpeedHQ/action@6b43a0cd438f6ca5ad26f9ed03ed159ed2df7da9 # v4.1.1
+        uses: CodSpeedHQ/action@c6574d0c2a990bca2842ce9af71549c5bfd7fbe0 # v4.2.1
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
           run: $CONDA/envs/test/bin/pytest --codspeed
--- a/tests/test_shards.py
+++ b/tests/test_shards.py
@@ -24,7 +24,7 @@ import requests.exceptions
 import zstandard
 from conda.base.context import context, reset_context
 from conda.core.subdir_data import SubdirData
-from conda.gateways.connection.session import get_session
+from conda.gateways.connection.session import CondaSession
 from conda.models.channel import Channel
 
 from conda_libmamba_solver import shards, shards_cache, shards_subset
@@ -142,7 +142,6 @@ def http_server_shards(xprocess, tmp_path_factory):
 
     malformed = {"follows_schema": False}
     bad_schema = zstandard.compress(msgpack.dumps(malformed))  # type: ignore
-    # XXX not-zstandard; not msgpack
     malformed_digest = hashlib.sha256(bad_schema).digest()
 
     (noarch / f"{malformed_digest.hex()}.msgpack.zst").write_bytes(bad_schema)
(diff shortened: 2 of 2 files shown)


conda_libmamba_solver/shards_subset.py
```diff
@@ -81,6 +94,119 @@ def in_shard(self, shardlike: ShardLike):
...
+    Fetch batches of shards from cache until in_queue sees None. Enqueue found
+    shards to shard_out_queue, and not found shards to network_out_queue.
+
+    When we see None on in_queue, we send None to both out queues and exit.
+    """
+    cache = cache.copy()
+
+    # should we send None to out_queues when done? Or just return?
+    running = True
+    while running and (batch := in_queue.get()) is not None:
```
dholth, line 113: Could this get / get_nowait() / combine batches / track running section be turned into a generator

Issue #749, opened 2025-10-20
Shards: precompute urljoin(urljoin(base_url, shards_base_url), (shard hash).hex() + ".msgpack.zst")


dholth:
### What happened?

urljoin is not fast. Compute the prefix for all shards so that an individual shard url is prefix + shard hash + ".msgpack.zst" instead of a urljoin() call.

Then it should be possible to use [Shards.shard_url()])https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/shards.py#L226) instead of e.g. [this workaround in index.py](https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/index.py#L596-L597)

Requires #750

PR #750, opened 2025-10-20
Shards: compatibility with rattler-index sharded repodata

### LINKED ISSUE #748: Support sharded repodata created by rattler
### Checklist

### What happened?

The experimental sharded repodata support does not work with shards (previously) created by rattler. This includes any channel on https://prefix.dev. 

We have identified a number of issues:

* zstd encoded files do not necessarily include the content-length in the frame header. This is not a requirement but can serve as an optimization when decoding. I think this should not be a requirement though. 
* prefix.dev specifically doesn't add trailing slashes in base urls like `shards_base_url`. We have since fixed this on prefix.dev but I still think the code should be robust enough to detect this scenario. I think in this context a base url like `https://blabla/something` should be assumed to have a trailing slash.
* rattler encoded byte arrays like hashes as an array of bytes instead of binary strings. rattler does supports decoding both forms. We have since [changed this in rattler](https://github.com/conda/rattler/pull/1755) (because binary strings are also smaller) but users that have used rattler-index previously will still have "conda incompatible" shards.

Although these issues have been largely mitigated on the prefix.dev servers there are still a large number of users that use rattler-index. Rattler-index also creates shards which will be incompatible with the current version of sharded repodata in this repository.

I hope these issues can be resolved to ensure maximum compatibility.

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

I used the following command to test the implementation:

```
CONDA_PLUGINS_USE_SHARDED_REPODATA=1 CONDA_DEBUG=1 pixi exec \
  -c main \
  -c conda-canary/label/dev \
  -s conda-canary/label/dev::conda-libmamba-solver=25.4.0.45 
  conda create --dry-run -c https://prefix.dev/conda-forge --override-channels conda
```

dholth:
### Description

Use `bytes(value).hex()` to handle `list[int]` or `bytes()` hashes.

Improve `base_url` handling by appending `/` if not empty, remove trailing filenames.

Give maximum output size to `zstandard.decompress()` so it will allow compressed data without the zstandard size header (added by zstandard when using one-shot instead of streaming compression APIs).

Fix #748

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/shards.py
+++ b/conda_libmamba_solver/shards.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
 
     from .shards_typing import PackageRecordDict, ShardDict
 
-ZSTD_MAX_SHARD_SIZE = 2**20 * 16  # maximum size necessary when compresed data has no size header
+ZSTD_MAX_SHARD_SIZE = 2**20 * 16  # maximum size necessary when compressed data has no size header
 # For reference, the largest shard "conda-forge/linux-64/vim" is 2608283 bytes
 # or < 2**19*5 decompressed (486155 bytes compressed); the index is 575219 bytes
 # decompressed (514039 bytes compressed) and is mostly uncompressible hash data.


conda_libmamba_solver/shards.py
```diff
@@ -134,7 +139,9 @@ def base_url(self) -> str:
 
         Note base_url can be a relative or an absolute url.
         """
-        return urljoin(self.url, self._base_url)
+        return urljoin(
+            urljoin(self.url, self._base_url), "."
+        )  # TODO do not call urljoin() when url, _base_url are unchanged
```
dholth, line 144: Addressed in https://github.com/conda/conda-libmamba-solver/issues/749

conda_libmamba_solver/shards.py
```diff
@@ -46,6 +46,8 @@
 
     from .shards_typing import PackageRecordDict, ShardDict
 
+ZSTD_MAX_SHARD_SIZE = 2**20 * 16  # maximum size necessary when compresed data has no size header
```
wolfv: just a minor typo (compresed vs compressed).

conda_libmamba_solver/shards.py
```diff
@@ -46,15 +46,20 @@
...
```
```suggestion
ZSTD_MAX_SHARD_SIZE = 2**20 * 16  # maximum size necessary when compressed data has no size header
```

travishathaway (approved):

Issue #754, opened 2025-09-25
Add cod-speed benchmark tests for repodata fetching


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

We need to add profiling for sharded repodata to measure it's performance versus the traditional method of repodata fetching.

### Why is this needed?

We need to make sure that this new feature is as performant as we think it is.

### What should happen?

- [ ] Setup benchmarks that focus on time elapsed comparing the old way of fetching repodata vs. sharded repodata fetching
- [ ] (Re)setup cod speed for conda-libmamba-solver

jaimergp:
FYI, there are some maybe-fitting tests in https://github.com/conda/conda-libmamba-solver/blob/main/tests/test_performance.py

PR #755, opened 2025-10-23
Adding Cod Speed + profiling tests

travishathaway:
### Description

fixes: #754 

### Checklist - did you ...

### DIFF
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -113,7 +113,6 @@ def test_reload_channels(tmp_path: Path):
     assert index.n_packages() == initial_count + 1
 
 
-@pytest.mark.benchmark
 @pytest.mark.parametrize(
     "load_type,load_channel", [("shard", "conda-forge-sharded"), ("repodata", "conda-forge")]
 )


.github/workflows/tests.yml
```diff
@@ -435,6 +423,7 @@ jobs:
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
           run: $CONDA/envs/test/bin/pytest --codspeed
+          mode: instrumentation
```
soapy1, line 426: It looks like in the codspeed UI it's suggesting to use [walltime mode](https://codspeed.io/docs/instruments/walltime) to measure these tests because of the system calls that happen. Is there a particular motivation for using the `instrumentation` mode here instead?

Image

tests/test_index.py
```diff
@@ -109,3 +111,38 @@ def test_reload_channels(tmp_path: Path):
...
+    conda_cli: CondaCLIFixture,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    """
+    Profile both loading methods and ensure they return the same data.
+    """
+    conda_cli("create", "--yes", "--prefix", str(tmp_path / "env"))
+    _, stderr, _ = conda_cli("clean", "--all", "--yes")
+
+    assert not stderr
```
soapy1, line 136: I think in this current form these long test setup steps are being included in the benchmarking. It would be better if these benchmarks just tested the repodata relevant code, opposed to this setup code. 

I think if you just extract these into fixtures it should be good. Otherwise, might want to adopt pedantic mode (https://codspeed.io/docs/reference/pytest-codspeed#pedantic-mode-advanced) that allows for seperating out setup and teardown steps before/after running code that is meant to be benchmarked.

.github/workflows/tests.yml
soapy1: hmmm, looking back at this, maybe if the test fixture setup is fixed to not measure the test setup steps then this will be resolved?

tests/test_shards.py
```diff
@@ -648,6 +648,51 @@ def test_batch_retrieve_from_cache(prepare_shards_test: None):
...
+    compressor = zstandard.ZstdCompressor(level=1)
+    for i in range(NUM_FAKE_SHARDS):
+        fake_shard = {f"foo{i}": "bar"}
+        annotated_shard = shards_cache.AnnotatedRawShard(
+            f"https://foo{i}",
+            f"foo{i}",
+            compressor.compress(msgpack.dumps(fake_shard)),  # type: ignore
+        )
+        cache.insert(annotated_shard)
+        fake_shards.append(annotated_shard)
```
soapy1, line 682: It looks like this is all setup for the tests. Can this be extracted into a fixture, or setup with pedantic mode so it doesn't get included in the benchmarking.

tests/test_shards.py
```diff
@@ -648,6 +648,51 @@ def test_batch_retrieve_from_cache(prepare_shards_test: None):
...
+
+    if retrieval_type == "retrieve_multiple":
+        retrieved = cache.retrieve_multiple([shard.url for shard in fake_shards])
+        assert len(retrieved) == NUM_FAKE_SHARDS
+
+    elif retrieval_type == "retrieve_single":
```
soapy1, line 701: It looks like the `retrieval_type` parameterization is just running 2 completely different code paths. So, I don't think there is too much to gain from parameterizing this test (for example, adding a new retrieval type will require adding a new branch to this if statement). 

It might be more clear to just split these up into 2 seperate tests with more descriptive names. Like `test_shard_cache_retrieve_multiple`,  `test_shard_cache_retrieve_single`.

.github/workflows/tests.yml
travishathaway: I chose "implementation" because I need to choose something. We could use wall time here. I could also using the context manager to more precisely measure the code were are profiling, specifically for the `test_load_channel_repo_info` test.

tests/test_shards.py
travishathaway: Okay, I'll break out the setup code into a separate fixture and refactor it so there's just two tests instead.

.github/workflows/tests.yml
jezdez: `walltime` is severely limited in time available for free (we're not paying for codspeed at the moment). When I enabled it a while ago, the time was gone in a day or so. 

From https://codspeed.io/pricing:

Image

docs/dev/index.md
```diff
@@ -3,11 +3,22 @@
...
 the following pages:
 
+- [How to set up your environment](setup)
+- [Typical development workflows](workflows)
+- [Implementation details](implementation)
+- [Profiling](profiling)
+- [Sharded repodata implementation](sharded)
+- [Future work](future-work)
+- [Code of conduct](code-of-conduct)
+- [CHANGELOG](changelog)
```
jezdez, line 13: <3

travishathaway: The current display for this is actually a complete mess. It currently shows all the parent and child links but with no indentation 😱 .

docs/dev/profiling.md
```diff
@@ -0,0 +1,24 @@
+# Profiling
```
jezdez: I'm not sure if "profiling" is the right term here, since that's only part of what is happening with the benchmark decorator, it does collect profile data of course, but also general metrics and I could see us group it under "performance testing"?

I'd also love for us to make better use of the `linux-memray` workflow step in the conda/conda repo for example, which uses `pytest-memray` to test for memory consumption. We don't do that anywhere here yet.

I know "performance" is a little more hand wavy, but we can add subsection to it like profiling with codespeed, memory use with memray and maybe more in the future?

.github/workflows/tests.yml
travishathaway: I'll just keep what I have then.

docs/dev/profiling.md
travishathaway: Sounds good. I will just change the header and the name of the file to "Performance testing" and `performance-testing.md` respectively.

travishathaway:
@soapy1 and @jezdez,

I have made several updates to the PR, and it's ready for another round of review.

tests/test_index.py
soapy1: Great change! Before, the benchmark results were:
|Benchmark|  Time (best)|Rel. StdDev|Run time| Iters|
|---------------------------------------------------------------------------------------|-------------------------|---------------|------------|-------|
│ test_load_channel_repo_info[shard-conda-forge-sharded] |16,349,500,516ns |      0.0% |  16.35s |    1 
│      test_load_channel_repo_info[repodata-conda-forge] | 16,988,191,719ns |        0.0% |   16.99s |     1 
│   test_shard_cache_multiple_profile[retrieve_multiple] |        283,364ns |      22.2% |   2.69s | 4,384 
│     test_shard_cache_multiple_profile[retrieve_single] |      697,319ns|       27.7% |   2.75s | 2,995 



Now that the tests don't include the setup:
|Benchmark|  Time (best)|Rel. StdDev|Run time| Iters|
|---------------------------------------------------------------------------------------|-------------------------|---------------|------------|-------|
│ test_load_channel_repo_info[shard-conda-forge-sharded] |   138,987,874ns |       15.0% |    2.96s |     18 
│      test_load_channel_repo_info[repodata-conda-forge] | 5,024,612,197ns |        0.0% |    5.02s |      1 
│   test_shard_cache_multiple_profile[retrieve_multiple] |         9,686ns |       18.6% |    3.00s | 29,260 
│     test_shard_cache_multiple_profile[retrieve_single] |       120,256ns |       25.8% |   3.00s |  7,797

PR #756, opened 2025-10-23
Pipelined sharded repodata traversal

### LINKED ISSUE #749: Shards: precompute urljoin(urljoin(base_url, shards_base_url), (shard hash).hex() + ".msgpack.zst")

### LINKED ISSUE #767: Keep decompression context in shard handling
> It is worthwhile to keep a decompression context. More pronounced in the relatively-faster-per-retrieval from-cache case. But dctx would need to be per-thread or to decompress in the main thread.
> 
> 
> 
> https://github.com/indygreg/python-zstandard/blob/main/zstandard/__init__.py#L211-L213 

 _Originally posted by @dholth in [#762](https://github.com/conda/conda-libmamba-solver/pull/762/files#r2487540545)_

dholth:
### Description

Workers for sqlite3 cache retrieval, network retrieval.

While shard traversal is happening, workers continuously empty queue of shards to fetch; try to fetch those shards in as large a batch as is available; and then dispatch negative results to the network worker or positive results to the queue of fetched shards.

The main thread only processes shards that are already in-memory, so that it can generate new batches of pending URLs without delay.

Fix https://github.com/conda/conda-libmamba-solver/issues/736

Fix #767 

Fix #749 

Continuation of https://github.com/conda/conda-libmamba-solver/pull/743, but branch is in conda/conda-libmamba-solver instead of own fork.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -373,7 +373,7 @@ jobs:
   # linux benchmarks
   linux-benchmarks:
     needs: changes
-    if: false and needs.changes.outputs.code == 'true'  # disable for now.
+    if: false # needs.changes.outputs.code == 'true'  # disable for now.
 
     runs-on: ubuntu-latest
     defaults:


tests/channel_testing/reposerver.py
```diff
@@ -349,7 +349,12 @@ def do_GET(self):
 
 PORT = args.port
 
-server = HTTPServer(("", PORT), handler)
+
+class HTTPServerBigQueue(HTTPServer):
+    request_queue_size = 128
```
dholth: This doesn't seem to solve the Windows GHA runner's refused connections error message in the network thread test. I was hoping that setting it to greater than the available test files might help.

conda_libmamba_solver/shards_subset.py
```diff
@@ -257,3 +407,160 @@ def build_repodata_subset(
...
+
+    def submit(node_id):
+        # this worker should only recieve network node_id's:
+        shardlike = shardlikes_by_url[node_id.channel]
+        if not isinstance(shardlike, Shards):
+            log.warning("network_fetch_thread got non-network shardlike")
+            return
+        session = shardlike.session
+        url = shardlikes_by_url[node_id.channel].shard_url(node_id.package)
+        return executor.submit(fetch, session, url, node_id)
```
dholth, line 555: This relies on the fact that the with statement doesn't create a scope. Will probably change it, but thought that was interesting.

conda_libmamba_solver/shards_subset.py
```diff
@@ -234,19 +255,150 @@ def shortest_bfs(self, root_packages):
...
+                shardlike = shardlikes_by_url[node_id.channel]
+                if node_id.package not in shardlike.visited:
+                    shardlike.visited[node_id.package] = shard
+                else:
+                    pass  # e.g. monolithic repodata is already visited
+
+                self.neighbors_2(pending, parent_node, shard)
+
+            enqueue_pending()
+            if not submitted:
```
dholth: This end criteria is buggy, and can be seen with `python -m conda create --dry-run --override-channels -c conda-forge-sharded vaex` with an empty shards cache. Should `neighbors_2` add to submitted, or should this be "if not submitted and not pending"?

conda_libmamba_solver/index.py
```diff
@@ -377,7 +391,13 @@ def _load_channel_repo_info_shards(
         """
         # make a subset of possible dependencies
         root_packages = (*self.in_state.installed.keys(), *self.in_state.requested)
-        channel_data = build_repodata_subset(root_packages, urls_to_channel)
+        algorithm = f"shortest_{_sharded_repodata_strategy()}"
+        print("Using", algorithm)
```
ryanskeith: Was this print intentional or debug?

conda_libmamba_solver/shards.py
```diff
@@ -560,13 +620,15 @@ def batch_retrieve_from_network(wanted: list[tuple[Shards, str, str]]):
...
 
-def fetch_channels(channels) -> dict[str, ShardBase]:
+def fetch_channels(channels: Iterable[Channel]) -> dict[str, ShardBase]:
     """
     Return a dict mapping of a channel URL to a `Shard` or `ShardLike` object.
 
     Attempt to fetch the sharded index first and then fall back to retrieving
     a traditional `repodata.json` file.
     """
+    # metaclass returns same channel, or casts to channel.
```
dholth, line 654: but the type checker doesn't understand this.

conda_libmamba_solver/shards_cache.py
```diff
@@ -68,6 +68,12 @@ def __init__(self, base: Path):
         self.base = base
         self.connect()
 
+    def copy(self):
```
ryanskeith, line 71: Aren't shards just sqlite bundles of data?  If they are, can't many processes read from them concurrently?  At least that was my understanding with sqlite databases.  Apologies for being uninformed here.

conda_libmamba_solver/shards_subset.py
```diff
@@ -234,26 +258,305 @@ def shortest_bfs(self, root_packages):
...
+
+        timeouts = 0
+
+        running = True
+        while running:
+            try:
+                new_shards = shard_out_queue.get(timeout=1)
+            except queue.Empty:
+                # print(f"Pending {len(pending)}, Submitted {len(submitted)}")
+                if all([x in self.nodes for x in submitted]):
```
ryanskeith: Should we add pending to this check as well like the other checks?

conda_libmamba_solver/shards_subset.py
```diff
@@ -234,26 +258,305 @@ def shortest_bfs(self, root_packages):
...
+                    # print("Done but submitted not empty")
+                    running = False
+                timeouts += 1
```
ryanskeith, line 324: timeouts are tracked but don't seemed to be used anymore.

conda_libmamba_solver/shards_subset.py
```diff
@@ -234,26 +258,305 @@ def shortest_bfs(self, root_packages):
...
+
+            enqueue_pending()
+            if not submitted and not pending:
+                break
+
+        cache_in_queue.put(None)
+        cache_thread.join()
+        network_thread.join()
+
+    def neighbors_2(self, pending: set[NodeId], parent_node: Node, shard):
```
ryanskeith: Since the signature is different then neighbors perhaps a name change is appropriate?

conda_libmamba_solver/shards_subset.py
```diff
@@ -85,18 +107,20 @@ class Node:
...
     def to_id(self) -> NodeId:
-        return NodeId(self.package, self.channel)
+        return NodeId(self.package, self.channel, self.shard_url)
 
 
 @dataclass(order=True, eq=True, frozen=True)
 class NodeId:
     package: str
     channel: str
+    shard_url: str = ""
```
ryanskeith, line 120: Is the NodeId creation in the `neighbors` function ok with this change?

ryanskeith (commented):
This looks like a great start. I saw a few things to polish.  I won't pretend I know a ton about threading in Python so I would suggest getting more eyes on this and/or do more testing.

conda_libmamba_solver/shards_cache.py
dholth: By default sqlite checks that it's being used from the creating thread. We could turn that off or make a new connection. I think the two connections will coordinate using filesystem locks if we are reading from cache and writing network results to cache at the same time, the same as if another process was using the same database.

conda_libmamba_solver/shards_subset.py
dholth: Yes, shard_url may be the default empty string.

conda_libmamba_solver/shards.py
```diff
@@ -314,6 +360,20 @@ def shard_url(self, package: str) -> str:
...
+        Return True if the given package's shard is in memory.
+        """
+        return package in self.visited
+
+    def visit_shard(self, package: str) -> ShardDict:
+        """
+        Return a shard that is already in memory and mark as visited.
+        """
+        shard = self.visited[package]
+        assert shard is not None
```
travishathaway:
```suggestion
```

I would prefer not having `assert` statements in the main part of the code. Let's keep these `assert` statements in `tests/`

conda_libmamba_solver/shards.py
```diff
@@ -142,6 +155,29 @@ def __contains__(self, package: str) -> bool:
...
+        """
+        Return shard URL for a given package. For monolithic repodata, should
+        not be fetched but is a unique identifier.
+
+        Raise KeyError if package is not in the index.
+        """
+        ...
+
+    @abc.abstractmethod
+    def shard_in_memory(self, package: str) -> bool:
```
dholth: not a big fan of these method names

conda_libmamba_solver/shards.py
```diff
@@ -221,6 +257,29 @@ def __repr__(self):
...
+        self.shards[package]
+        return f"{self.url}#{package}"
+
+    def shard_in_memory(self, package: str) -> bool:
+        """
+        Return True if the given package's shard is in memory.
+        """
+        return package in self.shards
+
+    def visit_shard(self, package: str) -> ShardDict:
```
dholth: We do need to traverse "shards" in monolithic repodata.json, but we could choose to send the original, full repodata.json into the solver straight from the disk cache after we've fetched all the real shards that are in the subset; instead of also converting all of these "from repodata.json" records to mamba PackageInfo.

conda_libmamba_solver/shards.py
```diff
@@ -314,6 +382,20 @@ def shard_url(self, package: str) -> str:
...
+        Return True if the given package's shard is in memory.
+        """
+        return package in self.visited
+
+    def visit_shard(self, package: str) -> ShardDict:
+        """
+        Return a shard that is already in memory and mark as visited.
+        """
+        shard = self.visited[package]
+        assert shard is not None
```
```suggestion
```

conda_libmamba_solver/shards.py
```diff
@@ -479,31 +561,38 @@ def fetch_shards_index(
...
-            # look for shards index
-            shards_index_url = f"{sd.url_w_subdir}/repodata_shards.msgpack.zst"
-            found = repodata_shards(shards_index_url, repo_cache)
-            cache_state.set_has_format("shards", True)
-            # this will also set state["refresh_ns"] = time.time_ns(); we could
-            # call cache.refresh() if we got a 304 instead:
-            repo_cache.save(found)
-
+        elif not repo_cache.stale():
+            # load from cache without network request
```
dholth, line 571: still missing offline mode?

conda_libmamba_solver/shards_subset.py
```diff
@@ -234,26 +259,333 @@ def shortest_bfs(self, root_packages):
...
+        shards_have = []
+        for node_id in pending:
+            # we should already have these nodes.
+            shardlike = shardlikes_by_url[node_id.channel]
+            if shardlike.shard_in_memory(node_id.package):  # for monolithic repodata
+                shards_have.append((node_id, shardlike.visit_shard(node_id.package)))
+            else:
+                if self.nodes[node_id].visited:
+                    print("skip visited")  # should not be reached
+                    continue
```
danyeaw, line 391: Hey @dholth, you mentioned having a challenge with tracking visited vs. unvisited nodes. What if we have the `in_flight` set of NodeIds checked here and skip them as well when nodes are in flight to prevent duplicate requests?

dholth: I think I solved it! This is the latest refactor (compare to Thursday morning). This line is in fact not reached. We now treat local, cached, and network nodes more similarly, they all spend time in pending and in-flight before their shard is processed, making it easier to know when we are finished.

conda_libmamba_solver/shards_subset.py
dholth: Changed

dholth:
This chart compares the "bfs" and "pipelined" strategies as run on my machine with cold, warm, and lukewarm cache. 
sharded-times-vaex
The sum of all the "bfs" durations is about 1.42x the sum of all the "pipelined" durations.

tests/test_shards_subset.py
```diff
@@ -130,9 +178,10 @@ def setup():
         return (subset,), {}
 
     def target(subset):
-        getattr(subset, f"shortest_{algorithm}")(scenario["packages"])
+        with _timer(""):
+            getattr(subset, f"shortest_{algorithm}")(scenario["packages"])
```
dholth: Would like to rename `shortest` to `reachability`

tests/test_shards.py
```diff
@@ -125,6 +125,30 @@ def prepare_shards_test(monkeypatch: pytest.MonkeyPatch):
     },
 }
 
+# This package depends on the
```
dholth, line 126: We could generate package(1-n), perhaps 100, and give each package a probability to depend on earlier packages in the series. More distant packages representing less likely dependencies. By using a fixed random number seed we could generate a predictable synthetic dependency graph.

conda_libmamba_solver/shards_cache.py
ryanskeith: Thank you for the explanation.

ryanskeith (commented):
Going over the code for maintainability, I think the threading code currently is on the harder side to maintain.   Or at least needs someone in the know of threading to maintain it.  My main experience has been with more embarrassing parallel workloads and this is a deeper level.  I do like though how the different jobs are clearly separated. Also, we don't have to use threading and still have a working option so risks are mitigated. The difficulty comes to me for reading the states.

conda_libmamba_solver/shards_subset.py
```diff
@@ -233,21 +266,179 @@ def shortest_bfs(self, root_packages):
...
+                    )
+                    log.debug("nodes: %d", len(self.nodes))
+                    log.debug("cache_thread.is_alive(): %s", cache_thread.is_alive())
+                    log.debug("network_thread.is_alive(): %s", network_thread.is_alive())
+                    # SimpleQueue doesn't have qsize(), but we can try
+                    try:
+                        log.debug("shard_out_queue.qsize(): %s", shard_out_queue.qsize())
+                    except AttributeError:
+                        pass  # SimpleQueue doesn't support qsize()
+                    if not pending and not in_flight:
```
dholth: end condition 1

conda_libmamba_solver/shards_subset.py
```diff
@@ -233,21 +266,179 @@ def shortest_bfs(self, root_packages):
...
+                    # add shard to appropriate ShardLike
+                    parent_node = self.nodes[node_id]
+                    shardlike = shardlikes_by_url[node_id.channel]
+                    shardlike.visited[node_id.package] = (
+                        shard  # would rather use the visit_shard (add shard?) method.
+                    )
+
+                    self.visit_node(pending, parent_node, shard_mentioned_packages(shard))
+
+                if not pending and not in_flight and not shutdown_initiated:
```
dholth: end condition 2. I would like to see if it's possible to consolidate the end conditions and avoid shutdown_initiated.

tests/test_shards.py
```diff
@@ -425,6 +480,7 @@ def test_shardlike_repr():
             "packages": {},
             "packages.conda": {},
             "info": {"base_url": "", "shards_base_url": "", "subdir": "noarch"},
+            "repodata_version": 1,
```
dholth, line 484:
```suggestion
            "repodata_version": 2,
```
However none of the limited amount of code that will process this repodata cares about the version number (we convert package records directly into libmamba objects)

conda_libmamba_solver/shards.py
jezdez:
```suggestion
    def loaded(self, package: str) -> bool:
```

Maybe that?

conda_libmamba_solver/shards_cache.py
```diff
@@ -61,16 +61,28 @@ class ShardCache:
...
         """
         self.base = base
         self.connect()
 
-    def connect(self):
+    def copy(self):
+        """
+        Copy cache with new connection. Useful for threads.
+        """
+        return ShardCache(self.base, create=False)
```
dholth, line 75: `sqlite3.OperationalError: disk I/O error` appeared in the thread, avoid redundant 'create table if not exists' in `copy()`.

PR #762, opened 2025-10-29
Implement "breadth-first search" as a graph traversal algorithm

travishathaway:
### Description

xref: #756 

This pull request is a direct result of my work on the issue linked to above. It implements a "breadth-first search" algorithm for tree traversal. So far, the performance improvements using this algorithm have been significant. In certain circumstances, fetching has improved by 300% (see: https://github.com/conda/conda-libmamba-solver/issues/736#issuecomment-3462077405).

Unfortunately, these performance differences cannot be seen with our current codspeed plan because it does not measure wall time. The efficiency gains we see here are happening because the traversal algorithm is collecting all the required nodes in fewer steps with larger groups. CPU time is the same because we're not doing any less computation, but the wall time is shorter because we're handling our network requests more efficiently.

This difference can be clearly seen when we fetch `python` and `pandas` as the root packages. The following is taken from a print statement I placed in `conda_libmamba_solver.shards.batch_retrieve_from_network` and represents the batch size of fetched shards.

`shortest_dijkstra`:

1. Packages fetched: 2
2. Packages fetched: 19
3. Packages fetched: 6
4. Packages fetched: 1
5. Packages fetched: 3
6. Packages fetched: 2
7. Packages fetched: 2
8. Packages fetched: 2
9. Packages fetched: 3
10. Packages fetched: 1
11. Packages fetched: 3
12. Packages fetched: 1 

`shortest_bfs`:
1. Packages fetched: 2
2. Packages fetched: 25
3. Packages fetched: 14
4. Packages fetched: 3
5. Packages fetched: 1

Above, the `shortest_dijkstra` completes fetching in 12 steps where `shortest_bfs` completes in 5 steps.

### Checklist - did you ...

### DIFF
--- a/tests/test_shards_subset.py
+++ b/tests/test_shards_subset.py
@@ -5,6 +5,7 @@ import pytest
 import pytest_codspeed
 from conda.common.compat import on_win
 from conda.models.channel import Channel
+from conda.testing.fixtures import CondaCLIFixture
 
 from conda_libmamba_solver.shards import fetch_channels
 from conda_libmamba_solver.shards_subset import RepodataSubset, build_repodata_subset
@@ -67,16 +68,19 @@ def codspeed_supported():
         return False
 
 
-def clean_cache(conda_cli):
+def clean_cache(conda_cli: CondaCLIFixture):
     """
     Clean cache and assert it completed without error except on Windows
     """
-    out, err, _ = conda_cli("clean", "--yes", "--all")
+    out, err, return_code = conda_cli("clean", "--yes", "--all")
 
-    # Windows CI runners cannot reliably remove this file, so we don't care about
-    # this assertion on that platform
+    # Windows CI runners cannot reliably remove this file, so we don't care
+    # about this assertion on that platform.
+
+    # "err" will include log.debug output on certain test runners, so we can't
+    # check it to determine whether there was an error.
     if not on_win:
-        assert not err
+        assert not return_code, "conda clean returned {return_code} != 0"
 
 
 @pytest.mark.skipif(not codspeed_supported(), reason="pytest-codspeed-version-4")
@@ -88,7 +92,7 @@ def clean_cache(conda_cli):
     ids=[scenario.get("name") for scenario in TESTING_SCENARIOS],
 )
 def test_traversal_algorithm_benchmarks(
-    conda_cli, benchmark, cache_state: str, algorithm: str, scenario: dict
+    conda_cli: CondaCLIFixture, benchmark, cache_state: str, algorithm: str, scenario: dict
 ):
     """
     Benchmark multiple traversal algorithms for retrieving repodata shards with


conda_libmamba_solver/shards.py
```diff
@@ -295,63 +331,56 @@ def fetch_shards(self, packages: Iterable[str]) -> dict[str, ShardDict]:
...
-                        p["name"]
-                        for p in (
-                            *result[fetch_result.package]["packages"].values(),
-                            *result[fetch_result.package]["packages.conda"].values(),
-                        )
-                    ]
-                    if set((fetch_result.package,)) != set(package_names):
-                        log.debug(
-                            "expected %s, actual %s", fetch_result.package, set(package_names)
-                        )
```
travishathaway, line 347: It looked like we were just doing this to log a debug statement, so I opted to remove it. This should rather be in our tests to ensure we are unpacking shards correctly.

conda_libmamba_solver/shards.py
```diff
@@ -105,7 +106,72 @@ def shard_mentioned_packages_2(shard: ShardDict) -> Iterable[str]:
...
+        """Return the names of all packages available in this shard collection."""
+        ...
+
+    @property
+    def base_url(self) -> str:
+        """
+        Return self.url joined with base_url from repodata, or self.url if no
+        base_url was present. Packages are found here.
+
+        Note base_url can be a relative or an absolute url.
```
jezdez, line 134:
```suggestion
        Note base_url can be a relative or an absolute url.
        The double urljoin ensures proper URL normalization:
        first join with _base_url, then with "." to add trailing slash
        if needed.
```

conda_libmamba_solver/shards.py
```diff
@@ -105,7 +106,72 @@ def shard_mentioned_packages_2(shard: ShardDict) -> Iterable[str]:
...
+        return package in self.package_names
+
+    @abc.abstractmethod
+    def fetch_shard(self, package: str) -> ShardDict:
+        """
+        Fetch an individual shard for the given package.
+
+        Raise KeyError if package is not in the index.
+        """
+        ...
```
jezdez, line 150:
```suggestion
    def fetch_shard(self, package: str) -> ShardDict:
        """
        Fetch an individual shard for the given package.

        Default implementation calls fetch_shards() with a single package.
        Subclasses may override for more efficient single-fetch operations.

        Raise KeyError if package is not in the index.
        """
        return self.fetch_shards([package])[package]

```

conda_libmamba_solver/shards.py
jezdez, line 327:
```suggestion
```

conda_libmamba_solver/shards.py
```diff
@@ -506,8 +542,10 @@ def batch_retrieve_from_cache(sharded: list[Shards], packages: list[str]):
 
 def batch_retrieve_from_network(wanted: list[tuple[Shards, str, str]]):
     """
-    Given a list of (Shards, package name, shard URL) tuples, group by Shards and call fetch_shards with a list of all urls for that Shard.
+    Given a list of (Shards, package name, shard URL) tuples, group by Shards and call fetch_shards
+    with a list of all URLs for that Shard.
     """
+    print(f"batch retrieve from network: {len(wanted)}")
```
```suggestion
```

conda_libmamba_solver/shards_subset.py
```diff
@@ -26,16 +30,33 @@
...
+```
+from conda.models.channel import Channel
+from conda_libmamba_solver.shards_subset import build_repodata_subset
+
+channel = Channel("conda-forge-sharded/linux-64")
+channel_data = build_repodata_subset(["python", "pandas"], [channel.url()])
+repodata = {}
+
+for url, shardlike in channel_data.items():
+    repodata[url] = channel_data.build_repodata()
```
jezdez:
```suggestion
for url in channel_data:
    repodata[url] = channel_data.build_repodata()
```

jezdez: Not sure if the example is not up-to-date, might be a thing we don't need as items

conda_libmamba_solver/shards_subset.py
```diff
@@ -127,22 +168,22 @@ def outgoing(self, node: Node):
         for n in self.neighbors(node):
             yield n, 1
 
-    def shortest(self, start_packages):
+    def shortest(self, root_packages):
```
jezdez:
```suggestion
    def shortest_dijkstra(self, root_packages):
```

I am not sure if you want to go there, but I thought if we're adding the ability to have multiple algorithms, might as well use the name?

travishathaway: I think that's a good idea. Otherwise it's less clear that we have those two implementations.

tests/test_shards.py
```diff
@@ -352,32 +368,6 @@ def test_shard_cache_multiple(tmp_path: Path):
...
-    # check that url's are in the url field, and package names are in the package name field
-    for url in extant:
-        if shard := cache.retrieve(url):
-            assert "://" in extant[url].url
-            assert "://" not in extant[url].package
-            for group in ("packages", "packages.conda"):
-                assert all(
-                    extant[url].package in (package["name"], "wrong_package_name")
-                    for package in shard[group].values()
-                ), f"Bad package name {extant[url].package}"
```
travishathaway, line 378: I initially discovered this test because it was causing failures in CI. When I examined it closer, it appears that it doesn't actually do what is described in the doc string. When I examined it even closer, it doesn't actually appear to do anything because that cache will always be empty at the beginning of this test unless a previous test has created cache.

For the reasons above, I chose to remove it entirely.

tests/test_shards_subset.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@cache
+def load_scenarios(scenario_filename: str) -> list[Scenario]:
+    """Load scenarios from tests/data folder and load them as `Scenario` objects."""
+    scenarios_file = Path(__file__).parent / "data" / scenario_filename
+
+    scenarios_raw = json.loads(scenarios_file.read_text())
+
+    return [Scenario(**scenario) for scenario in scenarios_raw.get("scenarios", [])]
```
kenodegard: I am not convinced storing the test cases in a separate JSON will make this easier to grok. We should definitely avoid this custom scaffolding, though as it adds unnecessary complexity. Something like the following accomplishes the same thing:


```python
PARAMETRIZE_SCENARIOS = pytest.mark.parametrize(
    "packages,prefetch_packages,channel",
    [
        pytest.param(
            scenario["packages"],
            scenario["prefetch_packages"],
            Channel(f"{scenario['channel']}/{scenario['platform']}",),
            id=scenario["name"],
        )
        for scenario in json.loads((Path(__file__).parent / "data" / "sharded_fetching_scenarios_basic.json").read_text()).get("scenarios", [])
    ],
)

...
@PARAMETRIZE_SCENARIOS
def test_traversal_algorithm_benchmarks(
    ...,
    packages: list[str],
    prefetch_packages: list[str],
    channel: Channel,
):
    ...
```

conda_libmamba_solver/shards.py
kenodegard: I disagree with this since `ShardLike` and `Shards` implement `fetch_shard`/`fetch_shards` differently

Since we define `ShardBase.fetch_shard` I was confused to see `ShardLike.fetch_shard` (especially without calling `super`)

For `ShardLike`:
- `fetch_shard` (singular) is the concrete implementation and
- `fetch_shards` (plural) is the helper method

For `Shards` we have the opposite:
- `fetch_shards` (plural) is the concrete implementation and
- `fetch_shard` (singular) is the helper method

conda_libmamba_solver/shards_subset.py
```diff
@@ -65,7 +89,7 @@ class Node:
     def to_id(self) -> NodeId:
         return NodeId(self.package, self.channel)
 
-    def in_shard(self, shardlike: ShardLike) -> bool:
+    def in_shard(self, shardlike: ShardBase) -> bool:
         return self.channel == shardlike.url
```
kenodegard: Unused.

conda_libmamba_solver/shards_subset.py
```diff
@@ -77,22 +101,38 @@ class NodeId:
     def __hash__(self):
         return hash((self.package, self.channel))
 
-    def in_shard(self, shardlike: ShardLike):
+    def in_shard(self, shardlike: ShardBase) -> bool:
         return self.channel == shardlike.url
```
kenodegard: Unused.

kenodegard (commented):
The benchmarks between `shortest_dijkstra` and `shortest_bfs` are effectively identical because the algorithms themselves are identical in this context (the weight/cost from a package to its immediate dependents is always 1).

The only implementation difference is using a heap queue versus a simple queue. Since all edges have the same weight/cost, both queue types act the same. We don't even need to use a queue, we can also make do with a plain list.

conda_libmamba_solver/shards.py
travishathaway: I'll switch this back to the way I had it before.

conda_libmamba_solver/shards.py
```diff
@@ -105,7 +106,78 @@ def shard_mentioned_packages_2(shard: ShardDict) -> Iterable[str]:
...
+    def __contains__(self, package: str) -> bool:
+        """Check if a package is available in this shard collection."""
+        return package in self.package_names
+
+    @abc.abstractmethod
+    def fetch_shard(self, package: str) -> ShardDict:
+        """
+        Fetch an individual shard for the given package.
+
+        Default implementation calls fetch_shards() with a single package.
```
```suggestion
```

No default implementation.

conda_libmamba_solver/shards.py
```diff
@@ -105,7 +106,78 @@ def shard_mentioned_packages_2(shard: ShardDict) -> Iterable[str]:
...
+        Update self.visited with all not-None packages.
+        """
+        ...
+
+    def build_repodata(self) -> RepodataDict:
+        """
+        Return monolithic repodata including all visited shards.
+        """
+        repodata = self.repodata_no_packages.copy()
+        repodata.update({"packages": {}, "packages.conda": {}})
```
dholth, line 164: We might have to sort the package names if we are worried about small changes affecting the solver. rattler doesn't sort package names in its repodata_shards.

conda_libmamba_solver/shards.py
```diff
@@ -215,26 +259,24 @@ def _shards_base_url(url, shards_base_url) -> str:
...
         self.shards_index = shards_index
         self.url = url
-        self.shards_cache = shards_cache
+        self.shards_cache = cache
 
-        # can we share a session for multiple subdir's of the same channel, or
-        # any time self.shards_base_url is similar to another Shards() instance?
-        self.session = get_session(self.shards_base_url)
+        # Use the channel's base URL to share session amongst subdir locations
+        channel_base_url = Channel(self.shards_base_url).base_url
```
dholth, line 271: Great find.
Maybe in the future we can change the way our sessions work, so that they are more shareable.

conda_libmamba_solver/shards.py
```diff
@@ -248,7 +290,7 @@ def __init__(
         self.visited: dict[str, ShardDict | None] = {}
 
         # https://github.com/conda/conda-index/pull/209 ensures that sharded
-        # repodata will always include base_url, even if it an empty string;
+        # repodata will always include base_url, even if it is an empty string;
         # this is also necessary for compatibility.
```
```suggestion
        # rattler/pixi require these keys.
```

conda_libmamba_solver/shards.py
```diff
@@ -281,77 +323,71 @@ def shard_url(self, package: str) -> str:
...
+    def _process_fetch_result(self, future, url, package, results):
+        """
+        Process a single fetched shard.
+        """
+        with conda_http_errors(url, package):
+            fetch_result = future.result()
+
+        # Decompress and save record
+        results[fetch_result.package] = msgpack.loads(
+            zstandard.decompress(
```
dholth, line 377: It is worthwhile to keep a decompression context. More pronounced in the relatively-faster-per-retrieval from-cache case. But dctx would need to be per-thread or to decompress in the main thread.

https://github.com/indygreg/python-zstandard/blob/main/zstandard/__init__.py#L211-L213

conda_libmamba_solver/shards.py
```diff
@@ -413,8 +449,15 @@ def fetch_shards_index(
...
     Check a SubdirData's URL for shards.
+
     Return shards index bytes from cache or network.
     Return None if not found; caller should fetch normal repodata.
+
+    TODO: If this function fails to retrieve the sharded repodata index file, it will
+          mark it is as not supporting this feature in cache. This can problematic
+          because sometimes server errors can happen which will lead it to wrongly
+          assuming the channel doesn't support sharding. We need to rethink our
+          logic for determining shard support.
```
dholth, line 453: IIRC the similar repodata gateway code has a set of error codes that mean "the file doesn't exist" including 404 of course, when checking for `repodata.json.zst` and others that mean "keep checking". If we mark it as missing incorrectly then we will still try again in 7 days.

conda_libmamba_solver/shards_subset.py
```diff
@@ -26,16 +30,33 @@
...
+```
+from conda.models.channel import Channel
+from conda_libmamba_solver.shards_subset import build_repodata_subset
+
+channel = Channel("conda-forge-sharded/linux-64")
+channel_data = build_repodata_subset(["python", "pandas"], [channel.url()])
+repodata = {}
+
+for url in channel_data:
+    repodata[url] = channel_data.build_repodata()
```
dholth, line 47: If we really do collect all `repodata[url]`, maybe we should try to do that in a way that each channel-subdir's repodata can be garbage collected immediately after being turned into libmamba-solver objects. Or Shards could pass individual package records to a callback without creating an in-memory repodata.json.

conda_libmamba_solver/shards_subset.py
```diff
@@ -26,16 +30,33 @@
...
+
 """
 
 from __future__ import annotations
 
 import heapq
-import json
 import logging
 import sys
+from collections import deque
```
dholth, line 59: *data structures* 🤤

conda_libmamba_solver/shards_subset.py
```diff
@@ -127,22 +162,22 @@ def outgoing(self, node: Node):
         for n in self.neighbors(node):
             yield n, 1
 
-    def shortest(self, start_packages):
+    def shortest_dijkstra(self, root_packages):
```
dholth, line 165: If it was proper un-buggy Dijkstra's, it should be breadth-first. "Dijkstra's algorithm finds the shortest path from a given source node to every other node." (Wikipedia)
It doesn't really matter what we call it.
I like continuing to compute the distance to each node for analysis purposes.

tests/test_shards_subset.py
```diff
@@ -0,0 +1,160 @@
...
+
+    warm:
+        all shards in the SQLite cache database
+
+    lukewarm:
+        only some of the shards are in the SQLite cache database
+    """
+    if cache_state == "warm":
+        # Clean cache just once for "warm"
+        clean_cache(conda_cli)
```
dholth, line 112: In another test we override conda's cache to a new temp directory, which is nicer instead of the user happening to clear their own cache while running tests locally.

conda_libmamba_solver/shards_subset.py
```diff
@@ -161,18 +196,64 @@ def initial_nodes():
...
+        """
+        self.nodes = dict(_nodes_from_packages(root_packages, self.shardlikes))
+
+        node_queue = deque(self.nodes.values())
+        sharded = [s for s in self.shardlikes if isinstance(s, Shards)]
 
+        while node_queue:
+            # Batch fetch all nodes at current level
+            to_retrieve = {node.package for node in node_queue if not node.visited}
+            if to_retrieve:
```
kenodegard, line 221: `node_queue` should only ever contain nodes/packages that have not been fetched yet, so checking for visited here is unnecessary?

conda_libmamba_solver/shards_subset.py
```diff
@@ -161,18 +196,64 @@ def initial_nodes():
...
+            # Process one level
+            level_size = len(node_queue)
+            for _ in range(level_size):
+                node = node_queue.popleft()
+                if node.visited:
+                    continue
+                node.visited = True
+
+                for next_node, _ in self.outgoing(node):
+                    if not next_node.visited:
```
kenodegard, line 234: `outgoing` and `neighbors` only return unvisited nodes, so checking for visited here is unnecessary?

dholth: I'd keep this because the check is cheap, and we could decide to remove the unvisited-only `if node_id not in self.nodes:` behavior. The "only unvisited neighbors" may prevent us from finding the true shortest path to each package? Which might be useful if we want to check our work.

conda_libmamba_solver/shards_subset.py
travishathaway: IIRC, someone told me dijstra's is just BFS without distance, so I figured they were different enough to warrant the label. Anyways, this gives us solid comparisons for our benchmarks. I decided to leave it here for a baseline comparison to make sure that we are actually iteratively making our software faster.

conda_libmamba_solver/shards_subset.py
travishathaway: I recently came across `pytest-memray` as a great profiling tool. Perhaps you could use that while experimenting with memory reduction strategies.

tests/test_shards_subset.py
```diff
@@ -0,0 +1,160 @@
...
+def clean_cache(conda_cli):
+    """
+    Clean cache and assert it completed without error except on Windows
+    """
+    out, err, _ = conda_cli("clean", "--yes", "--all")
+
+    # Windows CI runners cannot reliably remove this file, so we don't care about
+    # this assertion on that platform
+    if not on_win:
+        assert not err
```
dholth: err contains debugging output in the vscode test runner. Pull or cherry-pick https://github.com/dholth/conda-libmamba-solver/commit/9d86cec8852cf75319c9b6da0c9f88f26712946a for a fix, or I can try to push it here directly.

travishathaway: Please push directly to this branch to fix 🙏.

conda_libmamba_solver/shards.py
jezdez: Are you saying that without sorting this, there might be a difference in solution?

conda_libmamba_solver/shards.py
jezdez: I filed https://github.com/conda/conda-libmamba-solver/issues/767 to follow-up on it.

conda_libmamba_solver/shards.py
jezdez: @dholth Are you suggesting to follow the same logic for shards? How is rattler handling this?

conda_libmamba_solver/shards.py
dholth: I don't know but solvers are strange.

Issue #735, opened 2025-10-13
Discovery Summary: Update conda, conda-build CI as called by conda-libmamba-solver


dholth:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

CI does not pass reliably. Update conda-libmamba-solver and dependency CI's (conda, conda-build) to improve this.

conda-libmamba-solver checks out conda and runs conda tests.
Workflows are copied over from conda.

### What should happen?

Improve reliability of conda-libmamba-solver's CI by reviewing the test failures and looking for opportunities to make changes to the structure of tests.

jaimergp:
With `pytest-conda-solvers` this will be rendered redundant, so let's close this because we won't tackle it in any other way anyway.

PR #764, opened 2025-11-03
Update tests.yml to use reusable conda test suite workflow from branch

jezdez:
### Description

Refs #735

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,10 +78,10 @@ jobs:
     if: needs.changes.outputs.code == 'true'
 
     # Use reusable test workflow from conda/conda repo
-    # Pinned to commit: 195f731e1b9c1d44fb4bf7ad4d371cc9a1f4b13f (reusable-test-workflow branch) - includes CONDA_BUILD fix for Linux
+    # Pinned to commit: 458c66187 (reusable-test-workflow branch) - includes test fix for conda-libmamba-solver plugin settings
     # TODO: Update to a specific tag when conda releases a versioned tag
     # e.g., @v24.1.0 to pin to a specific conda version for reproducible tests
-    uses: conda/conda/.github/workflows/conda-test-suite.yml@195f731e1b9c1d44fb4bf7ad4d371cc9a1f4b13f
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -77,7 +77,11 @@ jobs:
     needs: changes
     if: needs.changes.outputs.code == 'true'
 
-    runs-on: windows-latest
+    # Use reusable test workflow from conda/conda repo
+    # Pinned to commit: 1ca8113c5d5aecaff86c994ddfd367dff91f2400 (reusable-test-workflow branch)
+    # TODO: Update to a specific tag when conda releases a versioned tag
```
jaimergp, line 82: Does `dependabot` pick these direct references too?

.github/workflows/tests.yml
```diff
@@ -91,142 +95,47 @@ jobs:
...
+            test-group: 2
+          - test-type: conda-libmamba-solver
+            test-group: 3
+    with:
+      runs_on: windows-latest
+      runner_os: Windows
+      shell: cmd /C call
+      checkout_conda: true
+      checkout_conda_path: conda
+      checkout_conda_ref: ''  # TODO: Set to a tag like 'v24.1.0' to pin conda version
```
jaimergp: We have been checking out `conda`'s repo to test against `main` so we pick up regressions before a release. This has been hindering development here though, so maybe this can be deferred to a cronjob that posts an issue if something breaks. A bit noisy with false positives, but less disruptive to ongoing development.

However, it will bite us when we find new deprecation warnings and so on. A compromise to be made.

.github/workflows/tests.yml
jezdez: I don't think so, no

jezdez: actually, it might

.github/workflows/tests.yml
jezdez: Something like 7bcf2e97ef933ad2e2192377d74737a17fbaeef6?

jezdez:
pre-commit.ci autofix

Issue #753, opened 2025-10-20
Adjust `conda update` message to frozen environments


marcoesters:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

The message that appears when a new version of `conda` is available should provide an accurate command to perform the update. A successful command will depend on whether the environment is frozen and whether `conda-self` is present in the `base` environment.

### Why is this needed?

This is needed to provide a better user experience. Users shouldn't encounter avoidable error messages when running a command that is being recommended to them.

### What should happen?

If the `base` environment is frozen, the message should read:

```
Please update conda by running

    $ conda update -n base -c {channel_name} conda --override-frozen
````
unless `conda-self` is present, in which case it should read:

```
Please update conda by running

    $ conda self update
```

stacynoland:
Actively working on PR for this issue.

PR #766, opened 2025-11-04
Adjust `conda update` message for frozen environments

stacynoland:
### Description

Fixes [#753](https://github.com/conda/conda-libmamba-solver/issues/753)

Tests are not included in PR because it requires an outdated version of `conda` in base, which would require a base environment change mid test run.

Completed tests locally using the following steps. Make sure `base` environment is not activated:

1. Clone base environment: `conda create -n base_test --clone base`

2. Activate test environment: `conda activate base_test`

3. Install `mspack-python` and tests/requirements.txt file: `conda install msgpack-python --file tests/requirements.txt`

4. Downgrade `conda`: `conda install conda=25.9.0`

5. Install local package: `python -m pip install -e . -vv --no-deps`

6. Activate `base_test` as `base` environment: `. <path to env>/etc/profile.d/conda.sh && conda activate`

7. Run `create` command and verify `conda update` message shows: `conda create -n unused --dry-run`

8. Add frozen marker file to environment: `touch $CONDA_PREFIX/conda-meta/frozen`

9. Run the same `create` command and confirm new message (below):
    `conda update -n base -c <channel_name> conda --override-frozen`

11. Install `conda-self`: `conda install conda-self --no-deps`

12. Run `create` command again and confirm new message (below):
    `conda self update`

### Checklist - did you ...

- [ ] ~Add / update necessary tests?~

### DIFF
--- a/conda_libmamba_solver/solver.py
+++ b/conda_libmamba_solver/solver.py
@@ -1030,8 +1030,10 @@ class LibMambaSolver(Solver):
         # once prefix data is lazy this might be a different situation
         current_conda_prefix_rec = None
         conda_self_installed = False
-        conda_frozen_file_found = os.path.exists(os.path.join(context.conda_prefix, PREFIX_FROZEN_FILE))
-        conda_meta_prefix_directory = os.path.join(context.conda_prefix, "conda-meta")
+        conda_frozen_file_found = os.path.exists(
+            os.path.join(context.root_prefix, PREFIX_FROZEN_FILE)
+        )
+        conda_meta_prefix_directory = os.path.join(context.root_prefix, "conda-meta")
         with suppress(OSError, ValueError):
             if os.path.lexists(conda_meta_prefix_directory):
                 for entry in os.scandir(conda_meta_prefix_directory):


conda_libmamba_solver/solver.py
```diff
@@ -1028,18 +1029,24 @@ def _notify_conda_outdated(
...
-                            current_conda_prefix_rec = PrefixRecord(**json.loads(f.read()))
-                        break
+                    if entry.is_file() and entry.name.endswith(".json"):
+                        package_name = entry.name.rsplit("-", 2)[0]
+                        if package_name == "conda":
+                            with open(entry.path) as f:
+                                current_conda_prefix_rec = PrefixRecord(**json.loads(f.read()))
+                        elif package_name == "conda-self":
+                            conda_self_installed = True
+                        else:
```
jaimergp: Do we need this `else` clause? It forces the loop to check only in the next file, but if we are unlucky and `os.scandir` returns either `conda` or `conda-self` last, then we will never check this condition. I think it's better to check in every loop.

stacynoland: If the last file is `conda` or `conda-self`, wouldn’t the loop break automatically after the `if…elif…else` statement? Moving the `else` clause so it is checked at the end of the each loop, instead of the next file in the loop, would expedite the `break` by one loop round. In the case where the last file is `conda` or `conda-self` though, the break should happen at the same point (end of the loop).

jaimergp: > If the last file is conda or conda-self, wouldn’t the loop break automatically after the if…elif…else statement?

Yes, you are right. 

In that case we can condense `else: if ...` into `elif ...`.

conda_libmamba_solver/solver.py
```diff
@@ -1028,18 +1029,24 @@ def _notify_conda_outdated(
         # manually check base prefix since `PrefixData(...).get("conda", None) is expensive
         # once prefix data is lazy this might be a different situation
         current_conda_prefix_rec = None
+        conda_self_installed = False
+        conda_frozen_file_found = os.path.exists(os.path.join(context.conda_prefix, PREFIX_FROZEN_FILE))
```
jezdez:
```suggestion
        conda_frozen_file_found = os.path.exists(os.path.join(context.root_prefix, PREFIX_FROZEN_FILE))
```

I think that'll consider if `ROOT_DIR` is used, and fallback to `context.conda_prefix` (aka `sys.prefix`).

conda_libmamba_solver/solver.py
```diff
@@ -1028,18 +1029,24 @@ def _notify_conda_outdated(
...
-                        and entry.name.rsplit("-", 2)[0] == "conda"
-                    ):
-                        with open(entry.path) as f:
-                            current_conda_prefix_rec = PrefixRecord(**json.loads(f.read()))
-                        break
+                    if entry.is_file() and entry.name.endswith(".json"):
+                        package_name = entry.name.rsplit("-", 2)[0]
+                        if package_name == "conda":
+                            with open(entry.path) as f:
+                                current_conda_prefix_rec = PrefixRecord(**json.loads(f.read()))
```
jezdez, line 1044: Oh fun discovering this code path, I'm not a fan of this casual file reading and JSON loading in the solver, two things that can go wrong (file system, and malformed JSON). Let's make this more defensive by a) moving that into a function and b) catch errors so this raises a nice message to users.

conda_libmamba_solver/solver.py
stacynoland: Shouldn't the `conda_meta_prefix_directory` on line 1034 also be updated with `context.root_prefix` too? I reused this line from the existing code, so it feels like both should be changed if this is considered more appropriate.

conda_libmamba_solver/solver.py
stacynoland: Most recent push resolves this.

conda_libmamba_solver/solver.py
stacynoland: Most recent push updates `context.conda_prefix` with `context.root_prefix`.

conda_libmamba_solver/solver.py
stacynoland: @jezdez I agree with you, but it feels a little out-of-scope for the PR and should be a new issue.

This JSON loading was already in the code and changing it doesn't pertain to the updated messaging for `conda-self`, which is the primary reason for this PR.

Also, I think some discussion needs to be had around if erroring out is the recommended approach or issuing the error while continuing the loop would be preferred.

dholth:
What is msgpack used for here?

stacynoland:
> What is msgpack used for here?

@dholth It was added for something to do with sharded repodata - nothing to do with this PR. I don't know details of the package or its use, but it was discussed on Slack in one of the channels recently as I had questions.

dholth:
@stacynoland duh, I am working on that project and we merged part of it to main. Thanks.

conda_libmamba_solver/solver.py
jaimergp: The need for this optimization goes away once we have [lazy PrefixData](https://github.com/conda/conda/pull/14811), hopefully in the 25.11 conda release.

jezdez: > This JSON loading was already in the code and changing it doesn't pertain to the updated messaging for conda-self, which is the primary reason for this PR.

We should fix this nevertheless.

jezdez (changes requested):

conda_libmamba_solver/solver.py
stacynoland: > > This JSON loading was already in the code and changing it doesn't pertain to the updated messaging for conda-self, which is the primary reason for this PR.
> 
> We should fix this nevertheless.

Then this should be a new issue that another PR can solve. The change needs more discussion around any error message handling and what that should look like, and that doesn't feel appropriate to handle that discussion here or in this PR which solves a completely unrelated issue.

jaimergp: I opened https://github.com/conda/conda-libmamba-solver/issues/784 to track this. @jezdez can you revisit your review? Thanks!

jaimergp (approved):

conda_libmamba_solver/solver.py
jezdez: Cheers, for the record, we've talked about this in person at the offsite and talked about https://github.com/conda/conda/pull/14811 in more depth. I'm reluctant to keep this code here, but assuming we fix the #784 in a fast follow-up PR, this can be merged.

PR #768, opened 2025-11-06
Remove shortest_ prefix from algorithm names

dholth:
### Description

Is nicer and can't run methods that don't begin with `shortest_`

Default to 10 parallel connections instead of checking requests' internals.

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/shards.py
+++ b/conda_libmamba_solver/shards.py
@@ -22,7 +22,7 @@ import msgpack
 import zstandard
 from conda.base.context import context
 from conda.core.subdir_data import SubdirData
-from conda.gateways.connection.session import CondaSession, get_session
+from conda.gateways.connection.session import get_session
 from conda.gateways.repodata import (
     _add_http_value_to_dict,
     conda_http_errors,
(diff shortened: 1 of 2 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -68,13 +68,14 @@ def _shards_connections() -> int:
     """
     if context.repodata_threads is not None:
         return context.repodata_threads
-    session = CondaSession()
+    # CondaSession() is expensive to create, we just want to find the default
+    # requests poolmanager size.
+    session = CondaSession.__base__() if CondaSession.__base__ else CondaSession()
```
travishathaway: Now that I'm looking at this again, I'm actually really confused why this level of object introspection is needed to get the number of concurrent requests allowable. We don't override this in conda, so I'm not sure why it's worth to go through so much trouble here.

Alternatively, we could just avoid `CondaSession` altogether and just create a normal `sessions.Session` object to get this information.

dholth: The idea was that conda might override the number in the future. On the other hand we could just be happy with SHARDS_CONNECTIONS_DEFAULT=10 or context.repodata_threads.

PR #792, opened 2025-11-26
Wait to fetch "fake shards" until we know we have some real shards.

dholth:
### Description

Skip shards code if there are none.

I think this works, but it could be more elegantly structured so that the network fetch was less intertwined with the construction of `ShardBase` objects.

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/shards.py
+++ b/conda_libmamba_solver/shards.py
@@ -532,18 +532,6 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
     return response_bytes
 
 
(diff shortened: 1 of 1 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -662,7 +662,7 @@ def fetch_channels(channels: Iterable[Channel | str]) -> dict[str, ShardBase]:
     # ),
     url_to_channel = dict(
         (channel_url, Channel(channel_url))
-        for channel in channels
+        for channel in channelsaoeu
```
```suggestion
        for channel in channels

jezdez (commented):
Looking good!

Issue #769, opened 2025-11-07
Shards: skip if no shards detected


dholth:
[index.py uses the shards code path if it is enabled](https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/index.py#L367)

Unfortunately it seems to be a bit slower than going straight through when there are no real sharded channels.

https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/shards.py#L577 fetches shards, but immediately fetches and loads monolithic repodata.json if a channel doesn't offer sharded repodata.

A simple solution would be to wait until all repodata_shards are fetched, and then decide whether to load the non-sharded channels as "ShardLike" channels.

The second time this runs, a cached "this channel doesn't offer repodata_shards.msgpack.zst" resolves without making network requests.

A more complicated solution might be to check if we have a cached "this channel had repodata_shards.msgpack.zst last time" and then either continue fetching shards and non-shards concurrently, or not.

danyeaw:
For the first implementation of this, we will start with the simple solution to return early.

PR #793, opened 2025-11-27
Add more robust detection for remote shard support

travishathaway:
### Description

closes: #769 

This pull request makes the detection of sharded repodata a little more intelligent. We first send `HEAD` requests to see if `repodata_shards.msgpack.zst` exists on the remote end and save that information. If any shards are detected, we perform shard retrieval, but no shards are detected, we fall back to our traditional repodata retrieval method.

I've added a new test in `tests/test_index.py` to more explicitly test the case when we use sharded and non-sharded together. This was inspired by a problem I ran into while developing that I wanted to make sure was covered where the requested package was in a non-sharded, but that package contained dependencies that were in a sharded channel.

Luckily, I think the existing tests should suffice to ensure this works, but I'm open to suggestions for more test cases to add.

My only hesitation about this pull request is the loose link between the sharded and non-sharded mock channels we have currently, but I think that should be taken care of in a separate issue. It would be nice to consolidate the mock channels in this repository so everything is in the same folder/module structure.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/builds-review.yaml
+++ b/.github/workflows/builds-review.yaml
@@ -39,7 +39,7 @@ jobs:
             }
       # Clean checkout of specific git ref needed for package metadata version
       # which needs env vars GIT_DESCRIBE_TAG and GIT_BUILD_STR:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           ref: ${{ github.ref }}
           clean: true
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,7 +22,7 @@ jobs:
       run:
         shell: bash -el {0}
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - uses: conda-incubator/setup-miniconda@835234971496cad1653abb28a638a281cf32541f # v3.2.0
         with:
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -23,7 +23,7 @@ jobs:
       GLOBAL: https://raw.githubusercontent.com/conda/infra/main/.github/global.yml
       LOCAL: .github/labels.yml
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - id: has_local
         uses: andstor/file-existence-action@076e0072799f4942c8bc574a82233e1e4d13e9d6 # v3.0.0
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -38,13 +38,13 @@ jobs:
 
     steps:
       - name: Checkout our source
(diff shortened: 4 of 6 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -480,7 +483,7 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
         headers["If-None-Match"] = str(etag)
     if last_modified:
         headers["If-Modified-Since"] = str(last_modified)
-    filename = "repodata_shards.msgpack.zst"
+    filename = SHARDS_INDEX_FILENAME
 
     with conda_http_errors(url, filename):
```
```suggestion
    with conda_http_errors(url, SHARDS_INDEX_FILENAME):
```

conda_libmamba_solver/shards.py
```diff
@@ -480,7 +483,7 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
         headers["If-None-Match"] = str(etag)
     if last_modified:
         headers["If-Modified-Since"] = str(last_modified)
-    filename = "repodata_shards.msgpack.zst"
+    filename = SHARDS_INDEX_FILENAME
```
```suggestion
```

conda_libmamba_solver/index.py
```diff
@@ -411,6 +417,40 @@ def _load_channel_repo_info_json(
...
+    def _is_sharded_remote_enabled(self, urls_to_channel: dict[str, Channel]) -> dict[str, bool]:
+        """
+        Check if sharded repodata is enabled by sending a HEAD request.
+
+        To do this, the presence of the "repodata_shards.msgpack.zst" file" is checked
+        """
+
+        def check(url: str) -> tuple[str, bool]:
+            session = get_session(url)
+            resp = session.head(f"{url}/{SHARDS_INDEX_FILENAME}")
```
jezdez: it's safer to use urljoin here instead of format strings

conda_libmamba_solver/index.py
```diff
@@ -364,8 +368,10 @@ def _load_channels(
 
         # Prefer sharded repodata loading if it's enabled
         if self.in_state and _is_sharded_repodata_enabled():
-            # TODO: It may be better to directly pass channel objects without URL encoding
-            return self._load_channel_repo_info_shards(urls_to_channel)
+            shard_support = self._is_sharded_remote_enabled(urls_to_channel)
+
+            if any(shard_support.values()):
```
jezdez: Ah!

tests/test_index.py
```diff
@@ -169,3 +169,43 @@ def index():
     index_helper = benchmark.pedantic(index, rounds=1)
 
     assert len(index_helper.repos) > 0
+
+
+def test_load_mix_shard_and_no_shard_channels(
+    http_server_auth_none,
```
jezdez, line 180: This fixture can't be found? I noticed that https://github.com/conda/conda-libmamba-solver/commit/f0894d987725a1f69e4b6adfca48866b50bdfe49#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R11-R17 does the fixture loading via imports. Please add `tests.channel_testing.helpers` to the `pytest_plugins` attribute so that pytest loads the `http_server_auth_none` fixture and fix the other fixtures while we're here.

travishathaway: That's a better a solution than the imports we currently have. Will do 🫡 .

conda_libmamba_solver/index.py
```diff
@@ -215,6 +221,65 @@ def _package_info_from_package_dict(
...
+        for url, channel in urls_to_channel.items():
+            subdir = SubdirData(channel)
+            try:
+                with subdir.repo_cache.cache_path_state.open("r") as fh:
+                    state = json.load(fh)
+                    is_supported = state.get("has_shards", {}).get("value", False)
+                    supported.append(is_supported)
+
+            except (OSError, json.JSONDecodeError, AttributeError) as exc:
+                log.debug(f"Unable to read state file for {channel}; reason: {exc}")
```
jaimergp, line 246: We usually log exceptions like this:

```suggestion
                log.debug("Unable to read state file for %s", channel, exc_info=exc)
```

Issue #710, opened 2025-08-11
Enable Offline Mode for Expired Cached Shards


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Provide the ability to use cached repodata_shards even when they have expired, specifically for offline usage scenarios.

### Why is this needed?

Users working in offline environments or with intermittent connectivity need access to previously cached data even if it's not guaranteed to be current. This improves the offline user experience significantly.

### What should happen?

Implement logic that can return cached repodata_shards when operating in offline mode, regardless of expiration status. The system should clearly indicate when using potentially stale data and provide appropriate warnings to users.

### Additional Context

Consider adding configuration options for offline behavior and provide clear feedback about the age and reliability of cached data being used.

PR #803, opened 2025-12-05
Enable offline mode with repodata shards

ryanskeith:
### Description

- Possible solution for issue #710
- Implemented a check to see if context is offline.  This is triggered by `--offline`, environment variable, etc.
  - If offline, check to see if there are cached shards and use those.  Warn if they are stale.
  - If offline and there is no cache, raise a RepoDataIsEmpty error
- Added tests

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,7 +34,7 @@ env:
   CONDA_LIBMAMBA_SOLVER_NO_CHANNELS_FROM_INSTALLED: true
   PYTEST_RERUN_FAILURES: 3
   # upstream conda tests that we want to skip
-  PYTEST_ADDOPTS: "-k (not test_get_index_lazy and not test_strict_resolve_get_reduced_index)"
+  PYTEST_ADDOPTS: "-k \"not test_get_index_lazy and not test_strict_resolve_get_reduced_index\""
   # See https://github.com/conda/conda/pull/13694
   # we can't break classic from here; no need to test it
   # those tests will still be available for local debugging if necessary


conda_libmamba_solver/shards.py
```diff
@@ -469,7 +469,33 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
...
+                )
+            else:
+                log.debug(
+                    "Using cached repodata_shards for %s (offline mode).", url
+                )
+            return cache.cache_path_shards.read_bytes()
+        else:
+            # In offline mode with no cache, signal that shards are not available.
+            # The caller (fetch_shards_index) catches RepodataIsEmpty and falls back to non-sharded repodata.
+            raise conda.gateways.repodata.RepodataIsEmpty(
```
ryanskeith: Perhaps a better error could be raised here.  Also, maybe change the status code to something that does not reflect HTTP like things.

conda_libmamba_solver/shards.py
```diff
@@ -469,7 +469,33 @@ def repodata_shards(url, cache: RepodataCache) -> bytes:
...
+    In offline mode, returns cached data even if expired. If no cache exists
+    in offline mode, raises RepodataIsEmpty to signal unavailability.
     """
+    # In offline mode, return cached data if available, even if expired
+    if context.offline:
+        if cache.cache_path_shards.exists():
+            # Check if cache is stale to provide appropriate warning message
+            is_stale = cache.stale()
+            if is_stale:
+                log.warning(
```
ryanskeith: Not sure if logs are helpful.  A user may not care if the cache is stale or not when explicitly asking for offline mode.

dholth: `conda create --offline --dry-run -p /tmp/conda conda`

No message, match that here.

conda_libmamba_solver/shards.py
```diff
@@ -571,6 +597,7 @@ def fetch_shards_index(
             # load from cache without network request
             shards_data = repo_cache.cache_path_shards.read_bytes()
 
+        # If we don't have shards_data yet, try fetching (repodata_shards handles offline mode)
```
dholth, line 586: repodata_shards() could be refactored, at the moment there isn't an elegant way for it to skip to returning the cached shard index. I think it should check cache_state.has_format("shards"), if that's False fall back to the non-shards offline support, if that's True load the cache without doing networking or checking for expiry.

https://github.com/conda/conda/pull/15471 begins trying to improve the recent binary cache API invented for sharded repodata so that there is a method call and not reading the info.json directly.

.github/workflows/tests.yml
```diff
@@ -597,6 +597,7 @@ jobs:
           python=${{ matrix.python-version }}
 
       - name: Conda Remove anaconda-auth
+        if: matrix.arch != 'osx-64'
```
dholth, line 600: thanks, we may also remove this conda remove again (after the tests have time to run, pass, and get merged) if the culprit is not these packages but the failing get_index_lazy test.

ryanskeith: It looks like we may have another problem with mac osx-64.  It seems it can't find it.  I also just noticed that conda only does arm atm.  Should we just drop osx-64 testing altogether?

dholth: I think conda-forge is still doing osx-64 so no we can't drop it.

PR #812, opened 2025-12-11
remove unused zstandard, requests dependencies

### LINKED ISSUE #811: shards: use compression.zstd instead of zstandard-python
### Checklist

### What is the idea?

conda-libmamba-solver should use the standard library `compression.zstd` when available, removing a dependency. Shards code uses `zstandard` currently.

dholth:
### Description

Instead of using `compression.zstd` in conda-libmamba-solver, it's all delegated to conda now.

Fix #811 

### Checklist - did you ...

### DIFF
--- a/news/811-compression-zstd
+++ b/news/811-compression-zstd
@@ -1,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Remove `requests`, `zstandard` dependencies; network access, shards are
+  delegated to conda. (#812)
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,8 +26,6 @@ dependencies = [
   "conda >=26.5",
   # "libmambapy >=2",
   "boltons >=23.0.0",
-  "requests >=2.28.0,<3",
-  "zstandard >=0.15"
 ]
 dynamic = [
   "version"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,8 +24,6 @@ requirements:
     - conda >=26.5
     - libmambapy >=2.0.0,!=2.6.2
     - boltons >=23.0.0
-    - requests >=2.28.0,<3
-    - zstandard >=0.15
 
 test:
   imports:


conda_libmamba_solver/shards_cache.py
```diff
@@ -128,16 +122,13 @@ def retrieve_multiple(self, urls: list[str]) -> dict[str, ShardDict | None]:
 
         # In one test reusing the context saves difference between .006s and .01s
         # We could make this a threadlocal.
-        dctx = zstandard.ZstdDecompressor()
+        dctx = ZstdDecompressor()
```
Rogdham: I would suggest using the `decompress` function directly instead of creating `dctx = ZstdDecompressor()` first:

1. `dctx` cannot be re-used
2. If the compressed data is composed of several Zstandard frames, you will only decompress the first one

conda_libmamba_solver/shards_subset.py
```diff
@@ -612,7 +611,7 @@ def network_fetch_thread(
         shardlikes: list of (network-only) shard index objects.
     """
     cache = cache.copy()
-    dctx = zstandard.ZstdDecompressor(max_window_size=ZSTD_MAX_SHARD_SIZE)
+    dctx = ZstdDecompressor()
```
Rogdham: Same here

conda_libmamba_solver/shards_cache.py
dholth: I see. Different than the other zstandard library.

.github/workflows/tests.yml
```diff
@@ -153,6 +153,7 @@ jobs:
           --file tests\requirements-s3.txt
           --file ..\conda-libmamba-solver\dev\requirements.txt
           ${{ matrix.test-type == 'conda-libmamba-solver' && '--file ..\\conda-libmamba-solver\\tests\\requirements.txt' || '' }}
+          ${{ matrix.python-version < '3.14' && 'backports.zstd' || '' }}
```
jaimergp:
```suggestion
          ${{ matrix.python-version < '3.14' && 'backports.zstd' || '' }}
```

Hm, are we relying of lexicographical sorting for this comparison? I guess since the lower bound is 3.10 anyway we don't have issues with `3.9 > 3.14`. Nice!

danyeaw (approved):

Issue #824, opened 2025-12-16
shards: keep channel ordering


dholth:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

A user noticed that `conda create -c local --dry-run python=3.14` could prefer the local channel, or could prefer the main channel.

sharded-repodata may be re-ordering channels based on the ordering of the `{channel_url: channel_data}` dict for example that could change based on download times. If that is the case, we might be able to keep track of the order outside of the shard download functions instead.

See [_set_repo_priorities](https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/index.py#L680)

```
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: file:///Users/user/envs/feedstock-tool/conda-bld/osx-arm64, prio: 3 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: file:///Users/user/envs/feedstock-tool/conda-bld/noarch, prio: 3 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/r/osx-arm64, prio: 2 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/main/noarch, prio: 2 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/r/noarch, prio: 2 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/main/osx-arm64, prio: 2 : 1
```

or

```
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/r/osx-arm64, prio: 3 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/main/noarch, prio: 3 : 1
DEBUG conda.conda_libmamba_solver.index:_set_repo_priorities(705): Channel: https://repo.anaconda.com/pkgs/r/noarch, prio: 3 : 1
[...]

PR #828, opened 2025-12-22
Keep channel ordering

soapy1:
### Description

fixes: https://github.com/conda/conda-libmamba-solver/issues/824

This PR does two main things:
* It refactors the test shard http servers to be able to create multiple test shard repositories. Note, in the current implementation they serve the same data
* Ensures that the channel order in the index is preserved

To test this out, you can try running the following command:
```
$ CONDA_CHANNELS=conda-forge CONDA_PLUGINS_USE_SHARDED_REPODATA=1 conda create --dry-run --solver=libmamba --channel=main python 
```

You would **expect** that this will try to install python from the main channel. For example, you can consider the output when shards are not enabled:
```
$ CONDA_CHANNELS=conda-forge CONDA_PLUGINS_USE_SHARDED_REPODATA=0 conda create --dry-run --solver=libmamba --channel=main python   

Channels:
 - main
 - conda-forge
Platform: linux-64
. . .

The following NEW packages will be INSTALLED:
. . .
  python             main/linux-64::python-3.14.2-h90eec9f_101_cp314 
. . .
```

* On the main branch, you can observe that the sharded repodata and non sharded repodata results are different
* On this branch, you should observe that they are consistent.

### Checklist - did you ...

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -33,8 +33,9 @@ from .channel_testing.helpers import (
 from .utils import conda_subprocess, write_env_config
 
 if TYPE_CHECKING:
-    from .test_shards import ShardFactory
     from conda.testing.fixtures import CondaCLIFixture, PathFactoryFixture, TmpEnvFixture
+
+    from .test_shards import ShardFactory
 
 DATA = Path(__file__).parent / "data"
 
--- a/tests/test_shards.py
+++ b/tests/test_shards.py
@@ -10,12 +10,12 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
+import tempfile
 import time
 import urllib.parse
 from contextlib import contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING, NamedTuple
-import tempfile
 
 import conda.gateways.repodata
 import msgpack


tests/test_shards.py
```diff
@@ -201,60 +201,85 @@ def shard_for_name(repodata, name):
 FAKE_SHARD = shard_for_name(FAKE_REPODATA, "foo")
 FAKE_SHARD_2 = shard_for_name(FAKE_REPODATA, "bar")
 
+class ShardFactory:
```
soapy1, line 206: This refactor can maybe be split into another PR?

tests/test_index.py
```diff
@@ -168,3 +169,38 @@ def index():
...
+    with patch(
+        "conda_libmamba_solver.index._is_sharded_repodata_enabled",
+        return_value=True,
+    ):
+        shard_enabled_index = LibMambaIndexHelper(**index_args)
+
+
+    # The expected output is that all of channel_one subdirs (noarch and linux-64) are
+    # ordered higher than channel_two subdirs.
+    expected_output_channels = [
```
soapy1, line 200: I think this is the right expectation here. That is, the channel order input to the index helper should be preserved.
Is it right that the platform specific channel gets injected in here?

tests/test_channels.py
```diff
@@ -482,3 +482,24 @@ def test_channels_are_percent_encoded(tmp_path):
     for repo in index.repos:
         assert "%20" in repo.url_no_cred
         assert "%20" in repo.url_w_cred
+
+
+@pytest.mark.integration
+def test_channel_ordering(conda_cli: CondaCLIFixture, monkeypatch: pytest.MonkeyPatch) -> None:
```
soapy1: should this integration test also use the http server shard fixture?

travishathaway: Yes, I think that would be better because we don't actually need to use `conda-forge` to prove channel order works, and not using local channels makes the tests take longer to run.

tests/test_index.py
travishathaway: Yes, I think it's correct. Needing to provide a subdir to the `Channel.from_url` method is just a little bit of awkwardness from the `Channel` API.

tests/test_shards.py
travishathaway: I think it would be okay to keep here, but I think it would be nice to have a doc string on this class explaining what it is and how it's meant to be used.

conda_libmamba_solver/index.py
```diff
@@ -370,7 +370,15 @@ def _load_channels(
...
-            return self._load_channel_repo_info_shards(urls_to_channel)
+
+            # Channels are not loaded in order from shards. Once we load the shards,
+            # reorder the list to match the intended channel order. If the channel
+            # is not in the urls_to_channel, we'll add it to the back
+            channel_repos_info = self._load_channel_repo_info_shards(urls_to_channel)
+            return sorted(
+                channel_repos_info,
+                key=lambda x: list(urls_to_channel.keys()).index(x.channel) or 0,
+            )
```
dholth, line 381: impressively short. no issues with the "expand linux-64 into noarch + linux-64"?

soapy1: This test case should be checked in `test_index.py::test_load_channels_order`. Could write more tests that are more focused on this expansion aspect. What do you think?

dholth: I would rather refactor the code so that the expansion happens at a higher level.

PR #837, opened 2026-01-09
Refactor sharded channel fetch

dholth:
### Description

Don't expand channels into `subdirs ⨯ channels` in shards code; leave it up to the caller `LibMambaIndexHelper`.

### Checklist - did you ...

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -495,15 +495,13 @@ def test_channel_ordering(
     # Setup two shard servers. server_one will have a small
     # delay in the response to mimic a slower response.
 
-    # Guarantee clean cache to avoid interference from previous tests
-    monkeypatch.setenv("CONDA_PKGS_DIRS", str(tmp_path))
-    monkeypatch.setenv("CONDA_PLUGINS_USE_SHARDED_REPODATA", "1")
-    reset_context()
-
     # This may have been noarch-only previously, but now it always tries
     # to fetch the current platform and fails.
-    server_one = http_server_shards_slow
-    server_two = http_server_shards
+    server_one = http_server_shards
+    server_two = http_server_shards_slow
+
+    # Guarantee clean cache to avoid interference from previous tests
+    monkeypatch.setenv("CONDA_PKGS_DIRS", str(tmp_path))
 
     monkeypatch.setenv("CONDA_CHANNELS", server_two)
     monkeypatch.setenv("CONDA_PLUGINS_USE_SHARDED_REPODATA", "1")
--- a/tests/test_shards.py
+++ b/tests/test_shards.py
@@ -310,7 +310,7 @@ class ShardFactory:
         )
 
         http = http_test_server.run_test_server(
-            str(shards_repository), finish_request_action=finish_request_action
+            str(self.root), finish_request_action=finish_request_action
         )
         self._http_servers.append(http)
 


conda_libmamba_solver/shards.py
```diff
@@ -657,29 +657,16 @@ def batch_retrieve_from_network(wanted: list[tuple[Shards, str, str]]):
...
-    # channels=[Channel(f"{load_channel}/linux-64")],
-    # subdirs=(
-    #     "noarch",
-    #     "linux-64",
-    # ),
-    url_to_channel = dict(
-        (channel_url, Channel(channel_url))
-        for channel in channels
-        for channel_url in channel.urls(True, context.subdirs)
-    )
```
dholth, line 680: I had to add a `noarch`-only option to the tests' `channels_to_dict()` function instead of always using `context.subdirs` as we do here or else some of the tests would try to fetch a non-existent "current platform" subdir from the test server. Hopefully this doesn't matter since we would have gone through the new expansion anyway in the non-testing code path.

conda_libmamba_solver/shards.py
```diff
@@ -718,4 +705,4 @@ def fetch_channels(channels: Iterable[Channel | str]) -> dict[str, ShardBase]:
             found = ShardLike(repodata_json, url)
             channel_data[channel_url] = found
 
-    return channel_data
+    return {url: channel for url, channel in channel_data.items() if channel is not None}
```
dholth, line 708: Maybe the filter by not-None is unnecessary, we should be guaranteed to fill them all in? Was thinking about the ongoing "return None if no shards" work...

conda_libmamba_solver/shards_subset.py
```diff
@@ -472,22 +472,18 @@ def drain_pending(
 
 def build_repodata_subset(
     root_packages: Iterable[str],
-    channels: dict[str, Channel] | list[Channel],
+    channels: dict[str, Channel],
```
dholth, line 475: Main point of this PR, always require the production channels dict instead of the testing-only list.

conda_libmamba_solver/index.py
```diff
@@ -458,6 +452,21 @@ def _channel_urls(self) -> dict[str, Channel]:
                     seen_noauth.add(url)
         return urls
 
+    @staticmethod
```
dholth, line 455: was inline in `_load_channels()`

conda_libmamba_solver/shards.py
soapy1: Reading the implementation above, it looks like `channel_data` is pre-populated with every url in the provided `url_to_channel` mapped to a `None` value.

Looking a the `fetch_shard_index` function, which populates the result of this function, it seems like `None` is a valid output.

I think, it's appropriate for this function to be dealing with `None` channels so that each caller of this function does not need to deal with sorting thru `None` channels. However, if it is important to account for each url that was passed into this function, it may be better to include the entries with a `None` channel.

conda_libmamba_solver/index.py
```diff
@@ -458,6 +452,21 @@ def _channel_urls(self) -> dict[str, Channel]:
...
+    def _encoded_urls_to_channels(urls_to_channel: dict[str, Channel]) -> dict[str, Channel]:
```
soapy1, line 456: Since this is used outside of this module by `tests/test_shards.py:channels_to_dict`, should the `_` be dropped?

```suggestion
    def encoded_urls_to_channels(urls_to_channel: dict[str, Channel]) -> dict[str, Channel]:
```

soapy1 (dismissed):
Nice one!

conda_libmamba_solver/index.py
dholth: IMO tests break the "private underscore" rule.

conda_libmamba_solver/shards.py
dholth: The futures_non_sharded side should fill in all the None's with a ShardLike, and we should have no None channels at exit.

dholth:
What's up with the test failures?

conda_libmamba_solver/shards.py
soapy1: ah ok, so in that case I think this function is producing a bad result if `None` in the output. Let's keep the filtering out `None` values. And when this ShardLike bits get implemented, the filtering can be removed? Maybe it would be helpful to leave a TODO/note here so that we don't forget about that.

PR #844, opened 2026-01-15
Set shards as unavailable if error code is (4xx and not 416) instead of on any error.

dholth:
### Description

We should fall back to `repodata.json`, but still look for shards next time, unless the `HTTPError` is most 4xx codes.

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/shards.py
+++ b/conda_libmamba_solver/shards.py
@@ -18,6 +18,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 from urllib.parse import urljoin
 
+import conda.exceptions
 import conda.gateways.repodata
 import msgpack
 import zstandard
@@ -30,7 +31,6 @@ from conda.gateways.repodata import (
 )
 from conda.models.channel import Channel
 from libmambapy.bindings import specs
-from requests import HTTPError
 
 from . import shards_cache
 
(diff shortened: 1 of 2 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -591,9 +600,21 @@ def fetch_shards_index(
...
                 # call cache.refresh() if we got a 304 instead:
                 repo_cache.save(shards_data)
-            except (HTTPError, conda.gateways.repodata.RepodataIsEmpty):
+            except conda.gateways.repodata.UnavailableInvalidChannel as err:
+                # repodata_shards converts HTTP errors to conda errors.
+                # fetch repodata.json / repodata.json.zst instead
+                if _is_http_error_most_400_codes(err.status_code):
+                    cache_state.set_has_format("shards", False)
+                repo_cache.refresh()
+            except conda.exceptions.CondaHTTPError as err:
```
dholth, line 609: This use of `CondaHTTPError` exception chaining is very unsatisfying, and it doesn't set the status code on itself but only formats it into a string.

Perhaps 'set_has_format("shards", False) on 404 only' is the correct answer which is handled by the UnavailableInvalidChannel exception, which is the superclass of RepodataIsEmpty.

jezdez: Oh, huh, that combination of exceptions are not super great, smells like a problem in the general networking/repodata loading stack to me, that we can't really fix fully here. Should we open a ticket upstream to have CondaHTTPError correctly contains the status code directly?

jezdez (approved):

conda_libmamba_solver/shards.py
dholth: [The conda_http_errors function is very complicated](https://github.com/conda/conda/blob/main/conda/gateways/repodata/__init__.py#L198) and was extracted out of [non-reusable error handling.](https://github.com/conda/conda/pull/12003/) in subdir_data.fetch_repodata_remote_request.

I think it would make sense to forget about the "most 400 error codes" logic and trap `UnavailableInvalidChannel` only.

At least our own http errors could make it easier to switch http backend libraries. What would a nicer error handling mechanism look like?

PR #845, opened 2026-01-15
Skip shard subset code if no shards detected.

### LINKED ISSUE #769: Shards: skip if no shards detected

dholth:
### Description

[Renamed](https://github.com/conda/conda-libmamba-solver/pull/792) branch. If a channel has no shards, fall back to the old code path.

Close sqlite3 connections explicitly to avoid a Python 3.13+ warning (split off into separate PR)

Fix #769

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,7 +146,6 @@ jobs:
           --file tests\requirements-${{ runner.os }}.txt
           --file tests\requirements-ci.txt
           --file tests\requirements-s3.txt
-          --file tests\requirements-truststore.txt
           --file ..\conda-libmamba-solver\dev\requirements.txt
           ${{ matrix.test-type == 'conda-libmamba-solver' && '--file ..\\conda-libmamba-solver\\tests\\requirements.txt' || '' }}
           python=${{ matrix.python-version }}
@@ -311,7 +310,6 @@ jobs:
           --file tests/requirements-${{ runner.os }}.txt
           --file tests/requirements-ci.txt
           --file tests/requirements-s3.txt
-          --file tests/requirements-truststore.txt
           --file ../conda-libmamba-solver/dev/requirements.txt
           ${{ matrix.test-type == 'conda-libmamba-solver' && '--file ../conda-libmamba-solver/tests/requirements.txt' || '' }}
           python=${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -448,6 +450,10 @@ def _process_fetch_result(self, future, url, package, results):
         """
         Process a single fetched shard.
         """
+        # Fail early if no cache to store the result.
+        if self.shards_cache is None:
+            raise ValueError("self.shards_cache is None")
+
```
dholth, line 454: Moving away from always having to put a shards_cache (and sqlite3 session) on Shards(), this function is not called by the default 'pipelined' algorithm.

conda_libmamba_solver/index.py
```diff
@@ -361,31 +361,39 @@ def _load_channels(
...
-            return sorted(
-                channel_repos_info,
-                key=lambda x: list(urls_to_channel.keys()).index(x.channel) or 0,
-            )
+            if channel_repos_info is not None:
+                urls_to_channel_keys = list(urls_to_channel.keys())
+                # Searching for the index of Channel() in a list[str] of URLs
+                # works because curiously Channel('conda-forge/noarch') ==
+                # 'https://conda.anaconda.org/conda-forge/noarch. Should we
+                # search over .values() instead?
```
dholth: May be able to remove this as the code tries to preserve order in a different way, by copying the channels dict's keys before loading anything.

jezdez: https://github.com/conda/conda/pull/13418 also exists and seems related?

travishathaway (approved):

conda_libmamba_solver/index.py
dholth: This probably could have sorted by the str keys the whole time instead of relying on odd Channel() behavior, but the update provides a different (preserve channel order) solution.

PR #847, opened 2026-01-16
Close sqlite3 connections promptly to avoid warning.

### LINKED ISSUE #843: Shards can leave sqlite3 connections open, causing a warning (in tests).
### What happened?

Sharded repodata should close sqlite3 connections explicitly to avoid warnings introduced in Python 3.13.

Noticed in the tests.

dholth:
Refactor tests. Improve coverage.

### Description

Instead of keeping a long-lived sqlite3 connection on `Shards()`, add it only when doing work and close it when finished. A context manager on `ShardCache()` helps to manage the connection.

Refactor tests, move "subset" tests into `test_shards_subset.py` and increase test coverage.

Fix #843 

Split from #845 

### Checklist - did you ...

### DIFF
--- a/conda_libmamba_solver/shards.py
+++ b/conda_libmamba_solver/shards.py
@@ -17,6 +17,7 @@ from collections import defaultdict
 from typing import TYPE_CHECKING
 from urllib.parse import urljoin
 
+import conda.exceptions
 import conda.gateways.repodata
 import msgpack
 import zstandard
@@ -29,7 +30,6 @@ from conda.gateways.repodata import (
 )
 from conda.models.channel import Channel
 from libmambapy.bindings import specs
-from requests import HTTPError
 
 from . import shards_cache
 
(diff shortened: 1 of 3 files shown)


conda_libmamba_solver/shards.py
```diff
@@ -314,20 +313,32 @@ class Shards(ShardBase):
     Handle repodata_shards.msgpack.zst and individual per-package shards.
     """
 
-    # cache for shards_base_url()
```
dholth, line 317: This was only relevant for the tests, now the test is a little more complex and the class is simpler.

conda_libmamba_solver/shards.py
```diff
@@ -657,25 +654,26 @@ def batch_retrieve_from_network(wanted: list[tuple[Shards, str, str]]):
...
     Returns:
-        A dict mapping channel URLs to `Shard` or `ShardLike` objects.
+        A dict mapping channel URLs to `Shard` or `ShardLike` objects. This dict
+        is return in the same order as the input `url_to_channel` dict.
     """
     # retain order from incoming dict:
     channel_data: dict[str, ShardBase | None] = {url: None for url in url_to_channel}
 
-    # share single disk cache for all Shards() instances
-    cache = shards_cache.ShardCache(Path(conda.gateways.repodata.create_cache_dir()))
```
dholth, line 693: Now the sqlite3 cache, connection is only added while doing traversal.

conda_libmamba_solver/shards_subset.py
```diff
@@ -607,7 +625,6 @@ def network_fetch_thread(
         cache: once shards are decoded they are stored in cache.
         shardlikes: list of (network-only) shard index objects.
     """
-    cache = cache.copy()
```
dholth, line 610: moved to with statement below

ryanskeith (dismissed):
These look good to me.  I am always a big fan of context managers.  We may want to make an issue for type annotations for this repository.  I noticed that there is random use throughout.

Also, we should fix the tests.

PR #856, opened 2026-01-23
Remove and retry on corrupted sqlite3 database

### LINKED ISSUE #823: shards: recover from corrupted repodata_shards.db
### Checklist

### What happened?

It is possible to have corruption in the sqlite3 database, for example, if someone writes `dd if=/dev/random of=repodata_shards.db bs=1k count=1`, the database will raise `sqlite3.DatabaseError: file is not a database` when `.execute()` is called on a connection; this would happen when we ensure the schema is available.

In my opinion conda might corrupt its own repodata_shards.db when multiple concurrent conda's run against a packages cache that is on a filesystem that does not support the locking used by sqlite3. Otherwise it could happen for a random reason.

It would also be possible for the database to have an unexpected schema, either because a future conda-libmamba-solver was in use or because it had been edited by the user. We should check that pragma user_version;
 == 0 and increment it if we do an (incompatible) schema change.

### Additional Context

It should recover. Work out the details about whether it clears the cache automatically, prompts the user to do so, renames the corrupted file or simply deletes it.

dholth:
### Description

Fix #823 

- AI assisted (tests)

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ env:
   # upstream conda tests that we want to skip
   PYTEST_ADDOPTS: "-k \"not test_get_index_lazy and not test_strict_resolve_get_reduced_index\""
   # for conda-libmamba-solver
-  PYTEST_CLS_ADDOPTS: "-m trouble -s" # focus on worst test(s)
+  PYTEST_CLS_ADDOPTS: "-m \"not slow and not benchmark and not trouble\""
   # See https://github.com/conda/conda/pull/13694
   # we can't break classic from here; no need to test it
   # those tests will still be available for local debugging if necessary


ryanskeith (commented):
I like the fix.  I would appreciate a comment explaining the if statement logic explaining behavior before and after 3.11.  Also, I am not sure if I should be concerned about the possibility of infinitely looping.  If you are not concerned, I won't be either.

conda_libmamba_solver/shards_cache.py
```diff
@@ -88,22 +88,31 @@ def copy(self):
...
+            # this schema will also get confused if we merge packages into a single
+            # shard, but the package name should be advisory.
+            self.conn.execute(
+                "CREATE TABLE IF NOT EXISTS shards ("
+                "url TEXT PRIMARY KEY, package TEXT, shard BLOB, "
+                "timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP)"
+            )
+        except sqlite3.DatabaseError as e:
+            has_errorcode = hasattr(e, "sqlite_errorcode")  # Added in Python 3.11
+            if retry and ((not has_errorcode) or (e.sqlite_errorcode == sqlite3.SQLITE_NOTADB)):
```
ryanskeith, line 118: Could we add a comment about always retrying this will always retry on Python <3.11 (first part) and only if it hits on SQLITE_NOTADB for Python 3.11+?  It took some time for me to grok this.

conda_libmamba_solver/shards_cache.py
```diff
@@ -88,22 +88,31 @@ def copy(self):
...
+                log.warning("%s '%s'; remove and retry.", dburi, e)
+                self.remove_cache()
+                return self.connect(create=create, retry=retry)
```
ryanskeith: Are you worried about infinitely looping here?

dholth: The flag is there specifically so we can pass False to the recursive call.

conda_libmamba_solver/shards_cache.py
ryanskeith: Is this reasonably?:
```
            # Backwards compatibility: Python < 3.11 can't distinguish error types,
            # so retry on any DatabaseError. Python 3.11+ only retries on SQLITE_NOTADB.
```

conda_libmamba_solver/shards_cache.py
ryanskeith: My fear is that retry is set to True by default.  So by default it will pass True repeatedly as retry=retry.

conda_libmamba_solver/shards_cache.py
dholth: I added the comment and `retry=False` on recursion.

PR #867, opened 2026-02-05
Use relative urljoin for non-HTTP schemes (s3://, file://, etc.)

### LINKED ISSUE #866: S3 channel URLs are corrupted due to urljoin not handling s3:// scheme
## Description

When using S3 channels (e.g., `s3://bucket-name/linux-64`), package URLs are corrupted to `.filename` instead of the full S3 URL.

## Root Cause

In `shards.py`, the `base_url` property uses Python's `urllib.parse.urljoin`:

```python
@property
def base_url(self) -> str:
    return urljoin(urljoin(self.url, self._base_url), ".")
```

However, `urljoin` doesn't recognize `s3://` as a valid URL scheme:

```python
>>> from urllib.parse import urljoin
>>> urljoin('https://repo.anaconda.com/pkgs/main/linux-64', '.')
'https://repo.anaconda.com/pkgs/main/'  # Correct!

>>> urljoin('s3://minio-s3-server/linux-64', '.')
'.'  # WRONG - should be 's3://minio-s3-server/'
```

Then in `index.py:621`, the URL is constructed via string concatenation:

```python
url=f"{base_url}{filename}"
```

When `base_url` is `.`, this produces `.zlib-1.2.11-h7b6447c_3.tar.bz2` instead of `s3://bucket/linux-64/zlib-1.2.11-h7b6447c_3.tar.bz2`.

## Error Message

```
conda.CondaMultiError: No connection adapters were found for '.zlib-1.2.11-h7b6447c_3.tar.bz2'
```

## Steps to Reproduce

1. Set up a local S3 server (e.g., Minio)
2. Create a conda channel on S3
3. Try to install a package from the S3 channel with libmamba solver

```bash
conda create -n test --channel s3://bucket-name zlib
```

## Workaround

Use the classic solver:
```bash
CONDA_SOLVER=classic conda create -n test --channel s3://bucket-name zlib
```

## Suggested Fix

The `base_url` property needs to handle URL schemes that `urljoin` doesn't recognize (like `s3://`, `ftp://`, etc.). One approach:

```python
from urllib.parse import urlparse, urljoin

@property
def base_url(self) -> str:
    parsed = urlparse(self.url)
    # urljoin doesn't handle non-http schemes properly
    if parsed.scheme and parsed.scheme not in ('http', 'https', ''):
        # For non-http schemes, manually construct the base URL
        if self._base_url:
            return urljoin(self.url, self._base_url)
        # Ensure trailing slash for proper concatenation
        return self.url.rstrip('/') + '/'
    return urljoin(urljoin(self.url, self._base_url), ".")
```

## Environment

- conda version: 26.1.0
- conda-libmamba-solver version: (current)
- Python: 3.13

jezdez:
## Summary

Python's `urllib.parse.urljoin` doesn't handle non-HTTP URL schemes properly. When joining an `s3://` URL with `"."` or `""`, it returns just `"."` instead of the expected directory URL. This caused package URLs to be malformed (e.g., `.zlib-1.2.11-h7b6447c_3.tar.bz2` instead of the full S3 URL).

This PR adds a `_safe_urljoin_with_slash()` helper that works correctly for all URL schemes by temporarily swapping non-HTTP schemes (like `s3://`, `file://`, `ftp://`) to `https://`, performing the standard `urljoin`, then restoring the original scheme. This ensures consistent `urljoin` semantics across all schemes without modifying global state.

## Changes

- Added `_safe_urljoin_with_slash()` helper in `shards.py` using a scheme-swap approach
- Updated `ShardBase.base_url` property to use the new helper
- Updated `_shards_base_url()` function to use the new helper
- Added parametrized tests covering HTTP, S3, file, and FTP URL schemes with both realistic file URLs and directory URLs
- Skipped flaky upstream test `test_reorder_channel_priority`

## Approach: scheme-swap

The implementation temporarily replaces non-HTTP schemes with `https://` before calling `urljoin`, then restores the original scheme on the result. This ensures all URL schemes follow standard `urljoin` path-resolution semantics:

- Last path segment without a trailing slash is treated as a filename and stripped (e.g., `s3://bucket/linux-64/repodata.zst` + `""` → `s3://bucket/linux-64/`)
- Trailing-slash directory URLs are preserved (e.g., `s3://bucket/linux-64/` + `""` → `s3://bucket/linux-64/`)

This is the same behavior HTTP URLs already had, now extended to all schemes.

## Alternative considered

Registering `s3` with `urllib.parse` globally (`uses_netloc.append("s3")`) was rejected because it modifies `urllib.parse` behavior for all Python code in the process, risks breaking other libraries, and doesn't cover all possible non-HTTP schemes.

Fixes #866

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ filterwarnings = [
   "ignore:conda.core.link.PrefixActions:DeprecationWarning:conda",
   "ignore:conda.core.link.PrefixActions:PendingDeprecationWarning:conda",
   "ignore:conda.base.constants.CONDA_PACKAGE_EXTENSIONS:PendingDeprecationWarning",
+  "ignore:conda.common.io.env_vars:PendingDeprecationWarning",
 ]
 markers = [
   "integration: integration tests that usually require an internet connect",


conda_libmamba_solver/shards.py
```diff
@@ -48,6 +48,53 @@
 
 SHARDS_CONNECTIONS_DEFAULT = 10
 ZSTD_MAX_SHARD_SIZE = 2**20 * 16  # maximum size necessary when compressed data has no size header
+
+# URL schemes that urljoin handles correctly
+_URLJOIN_SAFE_SCHEMES = frozenset(("http", "https", ""))
```
dholth: This isn't urljoin's longer list?

conda_libmamba_solver/shards.py
```diff
@@ -48,6 +48,53 @@
...
+
+
+def _safe_urljoin_with_slash(base_url: str, relative_url: str = "") -> str:
```
jaimergp, line 56: Wouldn't it be simpler to replace the scheme with `http` if it's something else, and then undo?

dholth: Good idea @jaimergp that is implemented here now.

.github/workflows/tests.yml
```diff
@@ -37,7 +37,7 @@ env:
   # Reruns on our tests
   PYTEST_RERUN_FAILURES_CLS: 0
   # upstream conda tests that we want to skip
-  PYTEST_ADDOPTS: "-k \"not test_get_index_lazy and not test_strict_resolve_get_reduced_index\""
+  PYTEST_ADDOPTS: "-k \"not test_get_index_lazy and not test_strict_resolve_get_reduced_index and not test_reorder_channel_priority\""
```
dholth, line 40: This test fails equally on main and this branch, and may have to do with newly-built packages on tested live channels.

conda_libmamba_solver/shards.py
jezdez: Good catch! https://github.com/conda/conda-libmamba-solver/pull/867/changes/bd5cd18373acbdb668dbd36ea7d95e4311a95447

PR #879, opened 2026-02-27
Replace IndexedSet with Sequence / tuple; additional typing tweaks

dholth:
### Description

See https://github.com/conda/conda/issues/15742

### Checklist - did you ...


conda_libmamba_solver/solver.py
```diff
@@ -139,7 +138,7 @@ def solve_final_state(
         ignore_pinned: bool | _Null = NULL,
         force_remove: bool | _Null = NULL,
         should_retry_solve: bool = False,
-    ) -> IndexedSet[PackageRecord]:
+    ) -> Sequence[PackageRecord]:
```
dholth, line 141: Base class only says (in an inaccurate docstring) that this returns a tuple of PackageRef

jaimergp: We can go with the more accurate:
```suggestion
    ) -> tuple[PackageRecord, ...]:
```

conda_libmamba_solver/state.py
```diff
@@ -472,12 +471,12 @@ def __init__(
         self.pins: dict[str, MatchSpec] = pins or {}
 
     @property
-    def current_solution(self) -> IndexedSet[PackageRecord]:
+    def current_solution(self) -> Sequence[PackageRecord]:
```
jaimergp, line 474:
```suggestion
    def current_solution(self) -> tuple[PackageRecord, ...]:
```

conda_libmamba_solver/state.py
```diff
@@ -519,7 +518,7 @@ def virtual_specs(self) -> dict[str, MatchSpec]:
         """
         return {name: spec for name, spec in self.specs.items() if name.startswith("__")}
 
-    def early_exit(self) -> IndexedSet[PackageRecord] | None:
+    def early_exit(self) -> Sequence[PackageRecord] | None:
```
jaimergp, line 521:
```suggestion
    def early_exit(self) -> tuple[PackageRecord, ...] | None:
```

dholth: I usually put the more generic type in annotations. `Sequence[]` would allow IndexedSet or tuple or list for example. Especially in the base class in conda.

conda_libmamba_solver/solver.py
dholth: tuple is more concrete, but Sequence is accurate and has a nicer "all elements of the same type" annotation without ...

PR #921, opened 2026-04-27
Cache `SolverInputState.installed` (B11)

jezdez:
### Description

`SolverInputState.installed` (`conda_libmamba_solver/state.py:220`) rebuilds `dict(sorted(prefix_data._prefix_records.items()))` on every property access. `_specs_to_request_jobs_add` in `solver.py:395` accesses it many times per solve, which on `conda update --all`-style workloads scales with the prefix size.

Phase-1 cProfile on W3 (synthetic 5,000-record prefix, `conda install --dry-run --no-deps tzdata`) showed:

    ncalls  tottime  cumtime  filename:lineno(function)
     10032   3.471   41.761   state.py:220(installed)

41.76 s out of 43.3 s solve time, or 96% of solve time at this scale.

Fix: cache the sorted result on the `SolverInputState` instance. The cache key includes the underlying `_prefix_records` mapping identity and `tuple(records)`, so the cached view is rebuilt when `PrefixData` reloads or when installed package names change, including same-size key swaps. Record values are expected to be stable for the lifetime of a solve, and in-place record mutations remain visible because the cached view holds the same record objects.

Part of the [conda-tempo Track B](https://github.com/jezdez/conda-tempo) performance work.

### Focused measurement

Cache-hit timing after installed-name tuple validation on local Darwin arm64, Python 3.10:

| N records | Rebuild sorted view | Cached access | Speedup |
|---:|---:|---:|---:|
| 1,000 | 93.0 us | 3.0 us | 30.7x |
| 5,000 | 497.9 us | 14.8 us | 33.5x |
| 10,000 | 982.9 us | 28.3 us | 34.7x |

The first access still pays the sort cost. Later accesses validate the installed-name tuple and reuse the cached sorted view, avoiding the repeated sorted dict rebuild that dominated the synthetic W3 profile. The earlier full-stack timing was removed because it covered a superseded combination of Track B changes rather than this PR in isolation.

Tracking issue: conda/conda#15969. Full research report: https://github.com/jezdez/conda-tempo/blob/main/track-b-transaction.md

### Checklist - did you ...

- [x] Add a file to the news directory ([using the template](https://github.com/conda/conda-libmamba-solver/blob/main/news/TEMPLATE)) for the next release's release notes?
- [x] Add / update outdated documentation? No user-facing docs change.

### DIFF
--- a/conda_libmamba_solver/state.py
+++ b/conda_libmamba_solver/state.py
@@ -224,7 +224,26 @@ class SolverInputState:
         can generate an equivalent ``MatchSpec`` object with ``.to_match_spec()``.
         Records are toposorted.
         """
-        return MappingProxyType(dict(sorted(self.prefix_data._prefix_records.items())))
+        # Every access previously rebuilt ``dict(sorted(...))`` from scratch,
+        # which is O(N log N) in the prefix size. ``_specs_to_request_jobs_add``
+        # reads this property many times per spec (see solver.py:442, 452,
+        # 458), producing an O(M * N log N) inner loop at solve time. For a
+        # 5 000-record prefix this was ~2.4 ms per access × ~10 000 accesses
+        # = 24 s just in this property.
+        #
+        # Cache the sorted view on the ``SolverInputState`` instance (the
+        # prefix is frozen for the life of one solve). Invalidate the cache
+        # if the underlying records mapping is swapped or mutated so the
+        # behaviour is identical to unconditional resorting. See #921.
+        records = self.prefix_data._prefix_records
+        cache_key = id(records), len(records)
+        cached_key = getattr(self, "_installed_cache_key", None)
+        if cached_key == cache_key:
+            return self._installed_cache
+        view = MappingProxyType(dict(sorted(records.items())))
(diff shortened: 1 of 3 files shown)


conda_libmamba_solver/state.py
```diff
@@ -224,7 +224,26 @@ def installed(self) -> dict[str, PackageRecord]:
...
+        # 458), producing an O(M * N log N) inner loop at solve time. For a
+        # 5 000-record prefix this was ~2.4 ms per access × ~10 000 accesses
+        # = 24 s just in this property.
+        #
+        # Cache the sorted view on the ``SolverInputState`` instance (the
+        # prefix is frozen for the life of one solve). Invalidate the cache
+        # if the underlying records mapping is swapped or mutated so the
+        # behaviour is identical to unconditional resorting. See #921.
+        records = self.prefix_data._prefix_records
+        cache_key = id(records), len(records)
```
jaimergp: This won't change if:

- A record is mutated
- A record is replaced by a different one

We could hash the keys only for some more protection, but not sure how expensive that is though

jaimergp (approved):

PR #926, opened 2026-04-30
Fix sqlite3 "database is locked" in pipelined shards cache

jezdez:
### Description

Fixes #924.

The pipelined shard traversal runs a `cache_fetch_thread` (reads) and a `network_fetch_thread` (writes) against separate `sqlite3.Connection`s to the same `repodata_shards.db`. With the default rollback journal, a reader cannot proceed while a writer holds the exclusive lock. Under CI load the 5 s default busy timeout expires and SQLite raises `sqlite3.OperationalError: database is locked`.

This PR applies two changes to `shards_cache.connect()`:

- **Enable WAL mode** (`PRAGMA journal_mode = WAL`), which allows readers to proceed against a snapshot while a writer is appending. The pragma is wrapped in a `try/except sqlite3.DatabaseError` so it degrades gracefully on locking-hostile filesystems (see #891) or corrupt databases.
- **Bump the connection timeout to 30 s** (from the 5 s default), giving the busy handler more headroom when WAL is unavailable.
- When WAL is confirmed active, also set `PRAGMA synchronous = NORMAL` (sufficient for a cache, avoids unnecessary `fsync`).

The dev script `dev/scripts/requests-fetch-all-shards.py` already used these same pragmas; this brings the production `connect()` in line.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda-libmamba-solver/blob/main/news/TEMPLATE)) for the next release's release notes?
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #927, opened 2026-04-30
Avoid sqlite3 lock contention by serializing access through the cache thread.

dholth:
### Description

Instead of giving the network thread a sqlite3 connection, add "store record" requests to the cache thread's input queue. The cache thread processes either "fetch record" or "store record" requests as they come in, and there is no database contention within a single conda process.

### Checklist - did you ...


jezdez:
I'd like to land #926 first as a quick fix since WAL is independently useful for any concurrent access to the cache. This can then build on top as a follow-up to remove the contention at the source. Alternatively, happy to fold this into #926 if you'd prefer shipping both together.

conda_libmamba_solver/shards_subset.py
```diff
@@ -575,6 +575,14 @@ def wrapper(in_queue, out_queue, *args, **kwargs):
     return wrapper
 
 
+class QueueCache:
+    def __init__(self, queue):
+        self.queue: Queue = queue
+
+    def insert(self, shard: AnnotatedRawShard):
```
jezdez, line 596: Instead of putting shards on the lookup queue and splitting with `isinstance`, a separate insert queue keeps the types clean:

```suggestion
class QueueCache:
    def __init__(self, insert_queue):
        self.insert_queue = insert_queue

    def insert(self, shard: AnnotatedRawShard):
        self.insert_queue.put(shard)
```

Then pass a new `SimpleQueue` as the insert queue and hand it to both `QueueCache` and `cache_fetch_thread`.

conda_libmamba_solver/shards_subset.py
```diff
@@ -597,6 +605,10 @@ def cache_fetch_thread(
     """
     with cache.copy() as cache:
         for node_ids in combine_batches_until_none(in_queue):
+            to_insert = [value for value in node_ids if isinstance(value, AnnotatedRawShard)]
+            for value in to_insert:
+                cache.insert(value)
+            node_ids = [value for value in node_ids if not isinstance(value, AnnotatedRawShard)]
```
jezdez: With the separate insert queue, the `isinstance` filtering goes away — just drain inserts before each batch of lookups:

```suggestion
    with cache.copy() as cache:
        for node_ids in combine_batches_until_none(in_queue):
            with suppress(queue.Empty):
                while True:
                    cache.insert(insert_queue.get_nowait())
            cached = cache.retrieve_multiple([node_id.shard_url for node_id in node_ids])
```

`cache_fetch_thread` just needs `insert_queue` as an extra argument.

jezdez (commented):
Nice idea to serialize all SQLite access through the cache thread. One suggestion to avoid the `isinstance` dance: use a separate insert queue instead of mixing types on the lookup queue.

conda_libmamba_solver/shards_subset.py
dholth: One thread can wait on one queue. That's why we use the queue for multiple purposes.

conda_libmamba_solver/shards_subset.py
dholth: Since it will only be a couple hundred, or a thousand in a bad case we should be able to afford it. My intuition is that having separate queues may cause shutdown issues. e.g. not guarantee that all items insert before cache thread shuts down.

conda_libmamba_solver/shards_subset.py
jezdez: Fair point — single blocking queue avoids the dispatcher needing to poll or coordinate readiness across two sources. Withdrawn.

conda_libmamba_solver/shards_subset.py
jezdez: Good catch on the shutdown ordering — a sentinel on the lookup queue would race with pending inserts on a second queue. Sticking with the single-queue dispatch makes sense.

dholth:
Will continue this in the conda repository.

dholth:
@jezdez if we really get frustrated, we could fall back to no-cache. Or the memory:// sqlite database! Are we closing the transaction correctly in our current write-to-cache?

conda_libmamba_solver/shards_subset.py
danyeaw, line 711:
```suggestion
)  # type: ignore[assign]
cache.insert(AnnotatedRawShard(url, node_id.package, data))
```
Remove stale comments.

danyeaw:
```suggestion
        cache_in_queue: SimpleQueue[list[NodeId] | list[AnnotatedRawShard] | None] = SimpleQueue()
```

danyeaw (approved):
Hey @dholth, this looks great, thanks for serializing access back down. I was thinking an additional unit test that creates a QueueCache, calls insert, and checks that the shard is in the output would have improve test coverage.

danyeaw (approved):

PR #939, opened 2026-05-06
Use shards subset code from conda

### LINKED ISSUE #908: Replace shard stack with imports from conda
### Checklist

### What is the idea?

After conda is updated with sharded repodata in `conda.gateways.repodata.shards`, conda-libmamba-solver stops hosting its own shards* modules and imports and calls conda’s gateway from index/plugin wiring instead.

### Why is this needed?

Prevents two implementations and keeps libmamba aligned with the canonical client owned by conda.

### What should happen?

- Local shard fetch/cache/subset logic is removed
- Implement plugin API for shards only

### Additional Context

Depends on https://github.com/conda/conda/issues/15906.

dholth:
### Description

Now that conda is getting a port of the conda-libmamba-solver shards fetcher, use the one that originates from conda.

I'd like to pass this in to the solver as a parameter, instead of having conda-libmamba-solver choose exactly which shards function it calls.

Fix #908

### Checklist - did you ...


conda_libmamba_solver/solver.py
```diff
@@ -70,6 +70,7 @@
         UpdateModifier,
     )
     from conda.common.path import PathType
+    from conda.gateways.shards import BuildRepodataSubset
```
dholth, line 73: implies requirement for conda >=26.5 although, since it's only in typing block it doesn't cause errors for older conda

travishathaway:
Hi @dholth,

I have one question about the implementation. Did you consider pulling what I call, "the ol' switcheroo"? By that I mean simply just swapping all of the modules verbatim to conda and then adding stubs to conda-libmamba-solver so that the imports continue to work for backwards compatibility purposes.

I was expecting to see a lot more removals in this pull request than there currently are. If we're not going to remove the sharded repodata implementation in this pull request, I'd like to see us at least come up with a plan for removing it in the future.

I'm not against the dependency injection strategy. I just wonder if we can't make this even simpler.

Also, I think this is being blocked by [this pull request](https://github.com/conda/conda/pull/15994), right? I wouldn't want to merge this while `CONDA_REF` is point at a feature branch.

dholth:
> Hi @dholth,
> 
> I have one question about the implementation. Did you consider pulling what I call, "the ol' switcheroo"? By that I mean simply just swapping all of the modules verbatim to conda and then adding stubs to conda-libmamba-solver so that the imports continue to work for backwards compatibility purposes.
> 
> I was expecting to see a lot more removals in this pull request than there currently are. If we're not going to remove the sharded repodata implementation in this pull request, I'd like to see us at least come up with a plan for removing it in the future.
> 
> I'm not against the dependency injection strategy. I just wonder if we can't make this even simpler.
> 
> Also, I think this is being blocked by [this pull request](https://github.com/conda/conda/pull/15994), right? I wouldn't want to merge this while `CONDA_REF` is point at a feature branch.

Since this implementation does not import new symbols from `conda` at runtime (only in the type checking block), we have simplified the release sequencing problem. We could release this today or after the necessary `conda` release without breaking anything and without requiring `conda > 26.5`.

(We do run conda-libmamba-solver's tests against the related CONDA_REF so that it will call us passing its own shards implementation, so that we know that both pieces work together.)

This isn't the main reason I've pursued this strategy but it is a side effect. I want to let `conda` choose which repodata subset function it passes to the solver.

We would certainly plan on removing `conda-libmamba-solver`'s shards implementation no later than the next release after this one.

.github/workflows/tests.yml
```diff
@@ -44,7 +44,7 @@ env:
   # we can't break classic from here; no need to test it
   # those tests will still be available for local debugging if necessary
   CONDA_TEST_SOLVERS: libmamba
-  CONDA_REF: main # pin conda version to latest for more stable tests
+  CONDA_REF: 15906-repodata-subset # branch with shards implementation
```
dholth:
```suggestion
  CONDA_REF: 15906-repodata-subset-g # branch with shards implementation
```

travishathaway:
@dholth,

That all makes sense. I think this is a fine approach then.

Are we waiting for `main` to be updated in [conda/conda](https://github.com/conda/conda) before we can merge this?

.github/workflows/tests.yml
```diff
@@ -44,7 +44,7 @@ env:
   # we can't break classic from here; no need to test it
   # those tests will still be available for local debugging if necessary
   CONDA_TEST_SOLVERS: libmamba
-  CONDA_REF: main # pin conda version to latest for more stable tests
+  CONDA_REF: 15906-repodata-subset-g # branch with shards implementation
```
jezdez:
```suggestion
  CONDA_REF: main # pin conda version to latest for more stable tests
```

jaimergp:
@dholth, is this ready? I think we can add the `conda>=26.5` lower bound now. Perhaps even remove the implementation and the tests? I don't think we ever promised bw-compat here, only in `conda`.

dholth:
This should be about ready modulo whatever the tests are doing. Now that this is releasing after conda, we can consider removing c-l-s's own implementation, but that might be easier in a second PR?

tests/test_solver.py
```diff
@@ -183,6 +189,8 @@ def test_update_from_latest_not_downgrade(
             "python",
         )
         update_python = PrefixData(prefix).get("python")
+        print(f"Check that {original_python} matches {update_python}")
+        print(update_python.__dict__)
```
danyeaw:
```suggestion
```

tests/test_solver.py
```diff
@@ -151,9 +151,13 @@ def test_determinism(tmpdir):
     assert len(set(installed_bokeh_versions)) == 1
 
 
+@pytest.mark.xfail(reason="conda-forge dependency changes?")
```
danyeaw, line 154: Hey @dholth, is this failure shards related?

conda_libmamba_solver/index.py
```diff
@@ -153,7 +151,8 @@ def _is_sharded_repodata_enabled():
     """
     Flag to see whether we should check for sharded repodata.
     """
-    return context.plugins.use_sharded_repodata is True  # type: ignore
+    # TODO this is referenced in the conda tests. Keeping here for compat.
+    return True
```
danyeaw, line 155: Hey @dholth, I see that conda has a comment in `tests/shards/conftest.py` about this function, but I don't see any real usage. What should our compatibility plan be for this and keeping around the now empty`test_plugin` module?

pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-  "conda >=26.1",
+  "conda >=26.5",
   # "libmambapy >=2",
   "boltons >=23.0.0",
   "msgpack >=1.1.1",
```
danyeaw:
```suggestion
```

Can we remove the msgpack dependency here and in `dev/requirements.txt` and `recipe/meta.yaml`?

danyeaw (commented):
Hi @dholth, I really like how you used dependency-injection so that conda owns the sharded repodata from now on, nice one!

I think for backwards compatibility, `plugins.use_sharded_repodata` is now ignored. Should we add a warning in conda, so that users are reminded to set `repodata_use_shards` instead? Is there anything else we need to be careful about regarding keeping things backwards compatible?

`docs/dev/sharded.md` still describes the deleted shards.py / shards_subset.py / shards_cache.py layout and tests/test_shards.py coverage model, so we should probably clean that up before merging.

tests/test_solver.py
dholth: We usually do remove debug prints, and I will here; but pytest hides stdout by default.

tests/test_solver.py
dholth: This is unrelated to shards. Jaime has a fix in a separate PR.

conda_libmamba_solver/index.py
dholth: I merged a PR that fixes this in conda.

danyeaw (approved):

PR #647, opened 2025-04-16
Fix `pytorch torchvision` GPU selection when CUDA is present

jaimergp:
### Description

Closes #646 

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda-libmamba-solver/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #962, opened 2026-06-08
Enforce constrains on virtual packages (conda/conda#10803)

carterbox:
### Description

The solver runs with `allow_uninstall=True`, so libsolv was free to "uninstall" a virtual package (e.g. __cuda) from the installed repo to dodge a run_constrained/constrains rule that references it -- silently ignoring a constraint like `__cuda>=13` on a host providing `__cuda==12`. Request virtual packages as (name-only) installs so they are pinned present and the constraint is enforced, mirroring conda's classic solver.

Companion to https://github.com/conda/conda/pull/16208
Part of fix for conda/conda#10803

When running the tests locally, `test_pytorch_gpu` from https://github.com/conda/conda-libmamba-solver/pull/647 failed. However, this test also did not pass for me when I checkout out the `main` branch, so it is unclear to me whether I have caused a regression or not. We will have to see whether the test passes in the CI.

### Checklist - did you ...


news/10803-virtual-package-constrains
```diff
@@ -0,0 +1,24 @@
...
+* <news item>
+
+### Bug fixes
+
+* Enforce `run_constrained`/`constrains` entries that reference virtual packages
+  (e.g. a `constrains` on `__cuda`). Virtual packages are now requested as
+  (name-only) installs so the solver -- which runs with `allow_uninstall=True` --
+  cannot drop them to dodge such a constraint. Previously a constraint like
+  `__cuda>=13` against a host providing `__cuda==12` was silently ignored.
+  (conda/conda#10803)
```
jaimergp:
```suggestion
  (conda/conda#10803 via #962)
```

jaimergp:
There's a regression in the CI pipeline that I'm fixing in https://github.com/conda/conda-libmamba-solver/pull/964.

carterbox:
FYI. I opened an issue in the mamba repo because micromamba has the same bug. It's unclear to me whether a fix in libmamba will make this PR obsolete.

https://github.com/mamba-org/mamba/issues/4311
https://github.com/mamba-org/mamba/pull/4312

conda_libmamba_solver/solver.py
```diff
@@ -420,6 +420,20 @@ def _specs_to_request_jobs(
             log.info("The solver will handle these requests:\n%s", json_str)
         return request_jobs
 
+    def _lock_virtual_packages(self, tasks: dict, in_state: SolverInputState) -> None:
+        """Pin the host's virtual packages present for a prefix solve.
+
+        Virtual packages (e.g. ``__cuda``) model immutable host state. The
+        solver runs with ``allow_uninstall=True``, so unless they are forced
+        present libsolv may drop a virtual package from the installed repo to
+        dodge a ``constrains``/``depends`` rule that references it -- silently
```
danyeaw:
```suggestion
        dodge a ``constrains`` rule that references it, and silently
```

Minor docstring nit, I don't think this impacts depends.

danyeaw (dismissed):
Thanks @carterbox, this looks like a great solution. Now we need to get the CI green to get this merged.

conda_libmamba_solver/solver.py
carterbox: Thanks for taking a look! 🙏🏼  I've revised this docstring.

conda_libmamba_solver/solver.py
```diff
@@ -425,6 +425,20 @@ def _specs_to_request_jobs(
...
+        Virtual packages (e.g. ``__cuda``) model immutable host state. The
+        solver runs with ``allow_uninstall=True``, so unless they are forced
+        to be present, libsolv may drop a virtual package from the installed repo to
+        dodge a ``constrains`` rule that references it (silently ignoring
+        ``__cuda>=13`` on a host providing ``__cuda==12`` for example).
+        Requesting them as (name-only) installs keeps them present so such
+        constraints are enforced. The conda-build path is intentionally
+        excluded.
+        """
+        tasks[Request.Install].extend(in_state.virtual.keys())
```
danyeaw:
```suggestion
        for record in in_state.virtual.values():
            tasks[Request.Freeze].append(record.to_match_spec().conda_build_form())
```

Hey @carterbox, now that CI is going again, there is a failure in `test_remove_globbed_package_names` from conda. That test is doing the following:

1. Create an env with zlib and ca-certificates from defaults
2. Run `conda remove *lib* --dry-run --solver=libmamba`
3. Expect success, with zlib in UNLINK

For the remove solve, the solver was building a job like Keep history packages, Remove glob matches, and then Install all virtual packages. That last step is not what we want, since we are trying to install things while simultaneously removing others. What if instead of `Request.Install`, we use `Request.Freeze`, which means keep the exact installed version.

conda_libmamba_solver/solver.py
```diff
@@ -425,6 +425,20 @@ def _specs_to_request_jobs(
...
 
+    def _lock_virtual_packages(self, tasks: dict, in_state: SolverInputState) -> None:
+        """Pin the host's virtual packages present for a prefix solve.
+
+        Virtual packages (e.g. ``__cuda``) model immutable host state. The
+        solver runs with ``allow_uninstall=True``, so unless they are forced
+        to be present, libsolv may drop a virtual package from the installed repo to
+        dodge a ``constrains`` rule that references it (silently ignoring
+        ``__cuda>=13`` on a host providing ``__cuda==12`` for example).
+        Requesting them as (name-only) installs keeps them present so such
```
carterbox:
```suggestion
        Freezing them at their current version keeps them present so such
```

conda_libmamba_solver/solver.py
carterbox: I had tried Freeze, but apparently I was missing some nuance related to `conda_build_form()`.

jaimergp:
We'll need https://github.com/conda/conda/pull/16332 for the `tests/core/test_solve.py::test_archspec_call[libmamba]` failures

jaimergp:
About `tests/cli/test_main_remove.py::test_remove_globbed_package_names[libmamba]`, the error does not reproduce on `osx-arm64`, but can be replicated there by mocking the subdir and virtual packages:

```
$ CONDA_OVERRIDE_GLIBC=2.28 CONDA_SUBDIR=linux-64 pytest "tests/cli/test_main_remove.py::test_remove_globbed_package_names[libmamba]"
```

There we get to the same exception:

```
conda_libmamba_solver.exceptions.LibMambaUnsatisfiableError: Encountered problems while solving:
  - conflicting requests
```

With some additional warnings worth mentioning:

```
WARNING conda.conda_libmamba_solver.solver:_prepare_problems_message(853): Failed to explain problems
Traceback (most recent call last):
  File "/Users/jrodriguez/devel/conda-libmamba-solver/conda_libmamba_solver/solver.py", line 850, in _prepare_problems_message
    explained_errors = unsolvable.explain_problems(db, problems_format_auto)
ValueError: Name of new element (_libgcc_mutex) does not match name of list (_libarchive_static_for_cph)
```

That [ValueError is raised internally by libmamba](https://github.com/mamba-org/mamba/blob/6bf836fde6a467751b17f49eeaf58e59a8c348d8/libmamba/src/solver/problems_graph.cpp#L798-L810). That code is not new (4 years ago), so not a recent regression. It was just not observed until now.

What may have changed is our solver requests. In this PR (need to enable `-vvv` in the CLI call for these to show up):

```
DEBUG conda.conda_libmamba_solver.solver:_solver_flags(397): Using solver flags:
{
  "allow_downgrade": true,
  "allow_uninstall": true,
  "force_reinstall": false,
  "keep_dependencies": true,
  "keep_user_specs": true,
  "order_request": true,
  "strict_repo_priority": false
}
INFO conda.conda_libmamba_solver.solver:_specs_to_request_jobs(425): The solver will handle these requests:
{
  "Keep": [
    "ca-certificates 2026.5.14 h06a4308_0",
    "zlib 1.3.2 h47b2149_0"
  ],
  "Remove": [
    "*lib*"
  ],
  "Freeze": [
    "__archspec 1 x86_64",
    "__conda 25.11.2.dev399 0",
    "__glibc 2.28 0",
    "__linux 0 0",
    "__unix 0 0"
  ]
}
```

On `main` it passes because there's no `Freeze` section, I guess.


```
DEBUG conda.conda_libmamba_solver.solver:_solver_flags(397): Using solver flags:
{
  "allow_downgrade": true,
  "allow_uninstall": true,
  "force_reinstall": false,
  "keep_dependencies": true,
  "keep_user_specs": true,
  "order_request": true,
  "strict_repo_priority": false
}
INFO conda.conda_libmamba_solver.solver:_specs_to_request_jobs(425): The solver will handle these requests:
{
  "Keep": [
    "ca-certificates 2026.5.14 h06a4308_0",
    "zlib 1.3.2 h47b2149_0"
  ],
  "Remove": [
    "*lib*"
  ]
}
```

jaimergp:
Apparently it works if we use `Install` for `(create|update|install)` and `Keep` for `remove`. 🤞 the whole test suite passes.

jaimergp:
`tests/test_repoquery.py::test_repoquery` failure on linux-aarch64 is unrelated (and fix is available at https://github.com/conda/conda-libmamba-solver/pull/964)

conda_libmamba_solver/solver.py
```diff
@@ -425,6 +425,21 @@ def _specs_to_request_jobs(
             log.info("The solver will handle these requests:\n%s", json_str)
         return request_jobs
 
+    def _lock_virtual_packages(self, tasks: dict, in_state: SolverInputState) -> None:
```
jezdez, line 428: This helper is used by the normal add/remove paths, but the conda-build request path below still skips it, so the same `allow_uninstall=True` escape remains there. I reproduced this by forcing `_called_from_conda_build()` and solving a local package constrained on `__cuda>=13` with `CONDA_OVERRIDE_CUDA=12.0`: current PR solves successfully, while adding the lock makes it raise the expected virtual-package conflict.

Could we also call it in `_specs_to_request_jobs_conda_build()` before returning?

```python
        self._lock_virtual_packages(tasks, in_state)
        return dict(tasks)
```

news/10803-virtual-package-constrains
```diff
@@ -0,0 +1,24 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Enforce `run_constrained`/`constrains` entries that reference virtual packages
+  (e.g. a `constrains` on `__cuda`). Virtual packages are now requested as
+  (name-only) installs so the solver -- which runs with `allow_uninstall=True` --
+  cannot drop them to dodge such a constraint. Previously a constraint like
```
jezdez: This wording still describes the earlier implementation as name-only installs, but the current code uses exact detected virtual package specs and `Keep` for remove solves per Jaime's later comment. Suggest wording this around keeping the detected virtual packages present instead.

```suggestion
  (e.g. a `constrains` on `__cuda`). Virtual packages are now kept present
  at their detected version/build in solver requests, so the solver -- which runs
  with `allow_uninstall=True` -- cannot drop them to dodge such a constraint.
```

jezdez (commented):
I left two inline suggestions: one to apply the virtual-package lock on the conda-build request path too, and one to bring the news entry in line with the current Install/Keep behavior.

conda_libmamba_solver/solver.py
carterbox: 2f125f4527156cfc021be2557b0f6b70e9282f32

jezdez (approved):

PR #978, opened 2026-06-25
Fix HTTP channel test isolation using conda's http_test_server

danyeaw:
### Description

I noticed that multiple CI runs were failing due to a test isolation problem on #973 with test runs like: https://github.com/conda/conda-libmamba-solver/actions/runs/28130872203/job/83307543630?pr=973#logs

This was caused by the http server tests changing the server config to add tokens and then not resetting the context afterwards.

To fix this, I imported conda's `http_test_server` and removed conda-libmamba-solver's xprocess based test server. This improves isolation as each test will get its own port number, and greatly improves performance because there is a large overhead of launching test servers with xprocess calls.

I also attempted to get rid of all subprocess calls, but it looks like conda is caching the resolved `defaults` multichannel process-wide. I added a comment about why the subprocess calls were kept.

Overall performance for these tests went from about 29s to 3s, a 8x speedup.

### Checklist - did you ...


soapy1 (approved):

danyeaw:
Thanks for the review @soapy1 🙌

PR #1009, opened 2026-07-30
`conda update --all` should update Python patch

### LINKED ISSUE #1008: `conda update --all` does not update Python within the current major.minor series
### Checklist

### What happened?

With the libmamba solver, `conda update --all` does **not** update Python to newer patch releases within the same major.minor series (e.g. `3.13.9` → `3.13.14`). The classic solver **does** under the same conditions.

#### Root cause

`SolverInputState.always_update` skips `python` under `UPDATE_ALL` (`pkg != "python"`). Python only gets a `python X.Y.*` pin and no Update request, so an installed patch already satisfying the pin stays put.

Classic includes Python in the `UPDATE_ALL` set and applies the major.minor business rule, which still allows patch upgrades. Exclusion came in with #270; intent was likely “don’t float across minors,” but it also blocks patches classic allows.

#### Expected

Parity with classic: update Python within current `X.Y.*` on `--update-all`; do not jump minors unless Python was requested explicitly.

#### Workarounds

- `conda install "python=3.13"`
- `conda update --all --solver=classic`
- Optionally add `python` to `aggressive_update_packages`

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

### Additional Context

- Related but different: https://github.com/conda/conda-rattler-solver/issues/100 (explicit `conda update python` minor jumps)
- Nearby: https://github.com/conda/conda-libmamba-solver/issues/156, https://github.com/conda/conda-libmamba-solver/issues/702
- Not listed in libmamba-vs-classic docs as an intentional deviation
- Fix direction: include `python` in `always_update` on `UPDATE_ALL`, keep the `X.Y.*` pin when not CLI-requested

kenodegard:
### Description

Resolves #1008 

### Checklist - did you ...


tests/test_state.py
```diff
@@ -37,3 +38,29 @@ def test_installed_cache_tracks_installed_names(tmp_path):
     )
 
     assert tuple(state.installed) == ("beta",)
+
+
+def test_update_all_includes_python_in_always_update(tmp_path):
```
danyeaw, line 46: It looks like this verifies membership in always allowed, I think the tests could cover a few more scenarios:
1. Patch upgrade is requested/allowed (3.13.9 → 3.13.14 style)
2. Minor is still blocked without an explicit python request
3. Pin + Update tasks are both emitted for python under UPDATE_ALL

danyeaw (commented):
Hey @kenodegard, thanks for this - this looks great and is the right fix. Just one comment about the tests being a bit thin.

ryanskeith (commented):
The change is sound.  I am still torn on the behavior though.  I think for safety though I think this is the better way, being `update all` seeks to preserve the minor Python version.  We should probably document behaviors we are seeking to enforce and which can be left up to the solvers.

danyeaw (approved):
Thanks @kenodegard!

Issue #108, opened 2023-01-13
Anaconda channels being fetched in addition to channels defined in .condarc 


andreostrovsky:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

When using `solver: libmamba`, conda fetches repodata for Anaconda.org channels whenever solving the environment:
```
pkgs/r/win-64
pkgs/msys2/noarch
pkgs/main/noarch
pkgs/msys2/win-64
pkgs/r/noarch
pkgs/main/win-64
```
despite these channels not being configured anywhere. My .condarc explicitly only lists conda-forge, accessed through my organization's jfrog artifactory. 
When using the classic solver, conda respects this. As does mamba. Yet when using the libmamba solver, it ignores this and pulls repodata for these channels regardless.
This is a security issue, as I do not want connections made to any sources outside my organization, and have explicitly configured conda as such.

I have verified: 
- In `prefix/pkgs/cache` the Anaconda.org channel repodata is in fact being downloaded.
- There are no packages installed in the environment from any of these channels.
- There is no mention of any of these channels in `prefix/pkgs/urls.txt`
- There is no mention of any of these channels in `prefix/conda-meta/history`
- I have run `conda clean --all`

### Conda Info

```shell
    active environment : base
   active env location : C:\Users\xxx\prefix
           shell level : 0
      user config file : C:\Users\xxx\.condarc
populated config files : C:\Users\xxx\.condarc
         conda version : 22.11.1
   conda-build version : not installed
        python version : 3.9.13.final.0
      virtual packages : __archspec=1=x86_64
                         __win=0=0
      base environment : C:\Users\xxx\prefix   (writable)
     conda av data dir : C:\Users\xxx\prefix\etc\conda
 conda av metadata url : None
          channel URLs : https://xxx:yyy@zzz/api/conda/remote-conda-forge-public/win-64
                         https://xxx:yyy@zzz/api/conda/remote-conda-forge-public/noarch
         package cache : C:\Users\xxx\prefix\pkgs
                         C:\Users\xxx\.conda\pkgs
                         C:\Users\xxx\AppData\local\conda\conda\pkgs
      envs directories : C:\Users\xxx\prefix\envs
                         C:\Users\xxx\.conda\envs
                         C:\Users\xxx\AppData\local\conda\conda\envs
              platform : win-64
[...]

PR #1027, opened 2026-08-17
Avoid using remote mirrors in tests

danyeaw:
### Description

CI was failing because `test_mirrors_do_not_leak_channels` depended on the Tsinghua Anaconda mirror, which now returns a 404 for one of the configured default channels. Remote mirrors also make this regression slow and unavailable to offline test runs.

This replaces the public mirror dependency with a fully local two-solve regression for #108:

- Install `test-package` from the shared local `mamba_repo` fixture.
- Rewrite the linked `conda-meta` record to advertise a separate, available local HTTP channel.
- Perform another solve without `--override-channels` and assert that the unconfigured channel receives no requests.

This preserves coverage for the behavior removed in #457 while testing the attempted network access directly. It also removes the no-op `CONDA_LIBMAMBA_SOLVER_NO_CHANNELS_FROM_INSTALLED` CI environment variable and a stale `index.py` docstring note.

### Checklist - did you ...

- [x] ~Add / update outdated documentation?~


jezdez (approved):

jezdez:
@jaimergp do you remember if these were intentionally mirroring from there?

jaimergp:
The gist here is that the mirrored packages still contained `conda.anaconda.org` in their `index.json` metadata, which would get in the way. Not sure if this test is relevant nowadays because I think we are now overwriting the channel information when the package is linked onto conda-meta. We were using the mirror out of convenience not necessity, since this is kind of easy to replicate offline too: we just need to have a local channel contain different `channel` information in the `index.json` files.

danyeaw:
Hey @jaimergp thanks for the context!
@jezdez With that extra info, I changed the test from creating a local mirror to testing against that stale index.json directly.

PR #1035, opened 2026-08-25
Run pytest-conda-solvers tests in our CI

### LINKED ISSUE #999: Add pytest-conda-solvers test suite to conda-libmamba-solver
### What is the idea?

Run pytest-conda-solvers YAML suite with `--conda-solver=libmamba` from this repo, without replacing the existing libmamba integration or upstream-conda suites (for now).

### Why is this needed?

Libmamba already has deep CLI/integration coverage and re-runs conda upstream with `CONDA_TEST_SOLVERS=libmamba`. The shared YAML suite is a declarative solve-case baseline that matches classic/rattler and is cheaper to interpret when porting or comparing solvers.

### What should happen?

- Checkout and install pytest-conda-solvers so `conda-solver-tests` is available.
- Add a CI job that runs the shared suite with `--conda-solver=libmamba` on Python 3.12+ matrix cells only.
- Document the local command (e.g. in `docs/dev/workflows.md`).
- Keep existing `tests/test_*.py` and the upstream conda checkout jobs as-is (for now)

### Additional Context

- https://github.com/conda-incubator/pytest-conda-solvers/issues/60
- This repo still CI-tests Python 3.10, make sure we gate the new job so it does not run on unsupported versions.

dholth:
### Description

Fix #999 

AI-assisted

### Checklist - did you ...


.github/workflows/pytest-conda-solvers.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Conda Info
+        run: python -m conda info --verbose
+
+      - name: Conda List
+        run: conda list --show-channel-urls
+
+      - name: Run solver tests
+        working-directory: pytest-conda-solvers
+        # same invocation as pytest-conda-solvers' own CI
+        run: python -m pytest --conda-solver=${{ env.CONDA_SOLVER }} tests/test_provenance.py -vv
```
dholth: ### 
```suggestion
        run: python -m pytest --conda-solver=${{ env.CONDA_SOLVER }} -vv
```
Maybe better to run all

ryanskeith (approved):
These changes look good to me.

jezdez (approved):

.github/workflows/pytest-conda-solvers.yml
```diff
@@ -0,0 +1,120 @@
...
+concurrency:
+  # Concurrency group that uses the workflow name and PR number if available
+  # or commit SHA as a fallback. If a new build is triggered under that
+  # concurrency group while a previous build is running it will be canceled.
+  # Repeated pushes to a PR will cancel all previous builds, while multiple
+  # merges to main will not cancel.
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+env:
```
jezdez, line 33: Could we make this test-only workflow read-only before it installs and executes the external checkout?

```suggestion
permissions:
  contents: read

env:
```

.github/workflows/pytest-conda-solvers.yml
```diff
@@ -0,0 +1,120 @@
...
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.12', '3.13', '3.14']
+
+    steps:
+      - name: Checkout conda-libmamba-solver
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          path: conda-libmamba-solver
```
jezdez, line 55: Please drop this checkout credential after fetch too. The later test code can otherwise reach it through this repository's Git configuration.

```suggestion
          path: conda-libmamba-solver
          persist-credentials: false
```

.github/workflows/pytest-conda-solvers.yml
```diff
@@ -0,0 +1,120 @@
...
+
+      - name: Checkout pytest-conda-solvers
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: dholth/pytest-conda-solvers
+          ref: patch-1
+          fetch-depth: 0
+          path: pytest-conda-solvers
```
jezdez: pytest-conda-solvers#97 has merged, so we can use the official repository and pin the reviewed tree. That also gives every matrix cell the same suite.

```suggestion
          repository: conda-incubator/pytest-conda-solvers
          ref: d757bfdb37c5235089c76ac4921f001713117b48
          fetch-depth: 0
          path: pytest-conda-solvers
          persist-credentials: false
```

dev/scripts/parse_workflow_runs.py
```diff
@@ -0,0 +1,107 @@
...
+    return name.replace("-", "_").upper()
+
+
+def substitute(text: str, env: dict, matrix_vars: set) -> str:
+    """Replace ``${{ env.X }}`` and ``${{ matrix.X }}`` expressions."""
+
+    def repl(match: re.Match) -> str:
+        kind, name = match.group(1), match.group(2)
+        if kind == "env":
+            return env.get(name, f"${{{shell_var(name)}}}")
```
jezdez: `env` values can be numbers or booleans. Parsing this repository's tests workflow currently crashes on `PYTEST_RERUN_FAILURES: 1`, so please normalize the value before returning it.

```suggestion
            return str(env.get(name, f"${{{shell_var(name)}}}"))
```

dev/scripts/parse_workflow_runs.py
```diff
@@ -0,0 +1,107 @@
...
+            run = step.get("run")
+            if run is None:
+                continue
+            if isinstance(run, list):
+                run = "\n".join(run)
+            run = str(run).rstrip("\n")
+
+            chunks.append(f"# {step.get('name', '(unnamed)')}")
+            if step.get("if"):
+                chunks.append(f"# if: {step['if']}")
```
jezdez: Could we fail clearly on step conditions until the parser can evaluate them? Right now it writes the condition as a comment and still executes the command.

```suggestion
            if "if" in step:
                raise ValueError(
                    f"step conditions are not supported: {step['if']}"
                )
```

dev/scripts/parse_workflow_runs.py
```diff
@@ -0,0 +1,107 @@
...
+            if step.get("working-directory"):
+                chunks.append(f"cd {step['working-directory']}")
```
jezdez, line 84: Each Actions step starts in its own working directory. This `cd` affects every later generated step, so please wrap the step in a subshell or restore the original directory.

jezdez (changes requested):
Thanks for putting this together, Daniel. The shared suite is a useful addition.

Two blockers need fixing before merge. The checked-out suite can access this repository's write-capable token, and the green matrix resolved `patch-1` at two different commits. I left direct suggestions for a read-only token and a canonical pinned checkout.

Could you also document the local command requested in #999? The helper should either preserve each step's conditions, directory, and environment values or fail clearly when it cannot.

dev/scripts/parse_workflow_runs.py
dholth: It was good enough to debug the action locally. I see that in the `.yml` only the last step contains `working-directory`

jezdez (approved):

PR #1047, opened 2026-09-11
Run solver test suites natively on Windows ARM64

jezdez:
### Description

Add Python 3.14 jobs on `windows-11-arm` to the main test workflow and the shared `pytest-conda-solvers` workflow. Both use native Python and conda-forge's new `libmambapy` 2.9.0 builds. Preserve the existing test matrix.

Use x64 Miniconda to bootstrap the native test environment until a native installer is available. Install dependencies with conda, using conda-forge first and defaults second with flexible channel priority. Defaults supplies native conda 26.7.2 and dependencies that are not yet available from conda-forge. Keep the existing editable source installs.

Run the full solver suite with its existing marker exclusions. Use the runner's Git and omit CodSpeed from the native job because its defaults package requires `pytest-benchmark<5.1`, conflicting with this suite's `>=5.1` requirement. Benchmarks are already excluded from these jobs.

Keep historical Python regression cases by solving for `win-64` where the requested packages were never published for ARM64. The solver process remains native, though subprocesses from those historical environments use x64 emulation. Use native Python and NumPy builds for the other installation tests, and check the test interpreter, installed package platforms, loaded libmambapy extension, and solver source location in the pytest header.

Related to conda/conda#16540 and conda/conda#15448.

### Checklist - did you ...

- [x] Add a file to the `news` directory for the next release's release notes?
- [x] ~~Add / update outdated documentation?~~

### DIFF
--- a/.github/workflows/pytest-conda-solvers.yml
+++ b/.github/workflows/pytest-conda-solvers.yml
@@ -46,6 +46,10 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.12', '3.13', '3.14']
+        include:
+          - os: windows-11-arm
+            python-version: '3.14'
+            subdir: win-arm64
 
     steps:
       - name: Checkout conda-libmamba-solver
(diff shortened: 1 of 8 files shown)


danyeaw (approved):
Thanks @jezdez!

PR #1053, opened 2026-09-15
Fall back to full repodata when an alternate file is unavailable

jezdez:
### Description

Conda-libmamba-solver uses `repodata.json` by default, but honors explicit `repodata_fns` settings from `--repodata-fn`, configuration files, or environment variables. When a requested alternate file is unavailable, `_fetch_one_repodata_json()` currently aborts rather than trying `repodata.json`.

Retry with full repodata in that case, matching `SubdirData._load()`. This does not change the default repodata filename. Keep the existing error for missing `noarch/repodata.json` and leave offline and cache-only behavior unchanged.

The bug surfaced in conda/conda#16664 because `test_remove_globbed_package_names` explicitly requested `current_repodata.json` with both solvers. Defaults includes `pkgs/r`, which has no `win-arm64` directory, so that request failed. The test now limits the flag to classic, where it was intended as a speed optimization. conda/conda#16664 no longer depends on this fix.

### Checklist - did you ...

- [x] ~~Add / update outdated documentation?~~


dholth:
I thought libmamba would never attempt `current_repodata.json`? Under which circumstances does this occur?

jezdez:
Yep, you're right, Daniel! The libmamba solver normally uses `repodata.json`. It keeps `current_repodata.json` when `repodata_fns` is explicitly configured, including via `--repodata-fn`.

Here, [the removal test](https://github.com/conda/conda/blob/91ba5a795f80741270b7a11beeea50cfd80b594d/tests/cli/test_main_remove.py#L79-L101) passed that flag to both solvers even though it was meant to speed up classic. I've limited it to classic in conda/conda#16664, so that PR no longer depends on this fix.

This is a separate fix for missing explicitly requested repodata files, including custom filenames. It doesn't change the default behavior. I've clarified that in the description too.

